Issue #3, opened 2020-08-24
Develop basic UI for Arcsi


gammaw:

gammaw:
@baxgas thanks & great! We'll close this one and open new issues as we go on with the arcsi front-end. If you have ideas, feel free to create new issues.

baxgas:
@gammaw I still need to work on it for this beta version, now that real data was available, there are some stuff that needs to be fixed.

pvj:
We need some more substantial front-end changes too, I think, apart from UI updates. Like for example: 

AJAX-based POST / PUT requests
Consolidate UX: top menu should make more sense (we don't need `add new items` for example, I think)
Actions such as `broadcast`, `archive`, `live` etc. shouldn't be on the `add` or `edit` pages, I think we will need a `dashboard` or `control panel` type view where the status of your show can be handled. 

@baxgas I trust you, you can basically do whatever you think is right but please think about these changes too that are more about the whole UX of the app than just the UI. 

Perhaps you know this too and this ticket is just about the UI in which case I jumped the gun a bit and I'm sorry.

gammaw:
We discussed in the IT meeting that we can test how it looks like now (locally) and if we like it we can merge and launch (close this one) and create specific new issues on demand. 

FYI @gammaw @pvj

gammaw:
@baxgas 
Couple of comments: 
Home:
- Long episode URLs still "overflow".
- Show headers with strange underline on ID part (why do we need the ID in the header?) 
- Links not visible on show cards.

Add item form: I was thinking that another alignment, e.g. where the different checkboxes are under each other, would be helpful. It's very important which checkbox is selected (because of Azura and DO) so it'd be better to have a more explicit visibility. 

![image](https://user-images.githubusercontent.com/28865986/97750779-be796780-1af1-11eb-9962-166caf36dbe0.png)

pvj:
Thanks for continuing the work and discussions around this! 

But wait 1 second! Before we move onto the specifics I'd like to have discussion with you guys on what should be presented and how the site look like in general. Like thinking radically drastically big for 1 hour or so.

It's my fault that such proposal is not even available to you guys. Sorry. And I would like to also not discredit or exclude you either. I just have an idea about how the site could be structured like and it is totally different from what is in the templates now. So I believe you might have such ideas too. 

[...]

PR #9, opened 2020-09-18
#3 basic ui for arcsi

pvj:
I checked this out yesterday on dev and man it looks soooo good already, I would like to merge this right now and continue from here.


gammaw (approved):
ZSÍR GUT!

Issue #27, opened 2020-10-28
Alphabetically sorted show names on add new episode form


gammaw:
The little brother of https://github.com/mmmnmnm/lahmacun_arcsi/issues/19

PR #28, opened 2020-10-29
Add sort function on show names

gammaw:
#27

I love lambda functions!


pvj (approved):
I love lambda too. Great addition, looks good to me!

One note for the future, I believe we will have to sort in multiple ways all over the place so ultimately I am hoping to implement sorts and filters on the API layer. What do you think of that? 

Meanwhile let's roll with this, perhaps add it to `edit_item` too?

gammaw:
> Meanwhile let's roll with this, perhaps add it to edit_item too?

@pvj good idea! I added it.

Issue #14, opened 2020-09-20
Auto suggest episode number


gammaw:
For example, based on the last know episode number (incremented by one).

pvj:
Yeah, this is UI/UX improvement, I think. 

The `show` object holds all the associated episodes so you could get the last one of them, increment by one, display it in the episode number field as the default value. 

cc @baxgas am I getting this right?

gammaw:
> The show object holds all the associated episodes so you could get the last one of them, increment by one, display it in the episode number field as the default value.

Good idea. I was thinking to remember the lastly added episode's number in the show object (database) but in case you're adding earlier episodes, it's not a good strategy. 

I will look into it.

gammaw:
Last commit implements a simple functionality: if no item number is specified, it looks for the highest existing number and increments it. 

The feature to suggest a number on the form would be a bit trickier because the show selector is on the same form. The proposed feature is simple but it does the job and it helps us well I think. 

The implementation may not be the most elegant one (it goes via the API and JSON-ing). A DB-only solution (via queries) would be certainly possible but this solution was easier for me now and since it doesn't really have a performance implication I just went for it.

pvj:
I get the general idea, and will review the PR shortly. I still think though that this should not be logic on the back-end.

gammaw:
@pvj closed the PR for various reasons: 
- I agree that a server-side implementation must be done via queries. 
- Currently, episode numbers are too chaotic and the increment logic would not be much help. 

I will icebox this issue and open a new one for a simply workaround: https://github.com/mmmnmnm/lahmacun_arcsi/issues/40

gammaw:
Stale + workaround issue exists, closing it.

PR #29, opened 2020-10-31
Simple auto increment logic for item number

gammaw:
#14

See more details in the issue. https://github.com/mmmnmnm/lahmacun_arcsi/issues/14

### DIFF
--- a/arcsi/api/item.py
+++ b/arcsi/api/item.py
@@ -15,6 +15,7 @@ from arcsi.handler.upload import DoArchive
 from arcsi.model import db
 from arcsi.model.item import Item
 from arcsi.model.show import Show
+from flask import current_app as app
 
 
 class ItemDetailsSchema(Schema):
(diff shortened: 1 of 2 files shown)


arcsi/api/item.py
```diff
@@ -91,9 +92,20 @@ def add_item():
     # work around ImmutableDict type
     item_metadata = request.form.to_dict()
     # TODO if we could send JSON payloads w/ ajax then this prevalidation isn't needed
+    show_id = int(item_metadata["shows"]) #shortcut in case the show id is needed in the next if statement
     item_metadata["shows"] = [
         {"id": item_metadata["shows"], "name": item_metadata["show_name"]}
     ]
+
+    if item_metadata["number"] == "":
+        show_request = requests.get(app.config["APP_BASE_URL"] + url_for("arcsi.view_show", id=show_id))
```
pvj, line 101: I've read your comment in issue #14 and while it is true it doesn't have perf indication with our current usage we will definitely have to write this over so why not try it right first? This feels bit unnecessary. In my opinion, we have the `Show` object available to instantiate and could query & filter that which returns only the latest episode only if it has one, all in one.

arcsi/api/item.py
```diff
@@ -91,9 +92,20 @@ def add_item():
...
+        show_dict = show_request.json()
+        if len(show_dict["items"])>0:
+            items_sorted = sorted(show_dict["items"], key=lambda k: k['number'], reverse = True)
```
pvj, line 104: 👍

arcsi/templates/item/add.html
```diff
@@ -30,7 +30,7 @@ <h2>{% block title %}Add episode to archive{% endblock %}</h2>
 </div>
 
   <div>
-    <label for="number">Episode number</label>
+    <label for="number">Episode number (+1 by def)</label>
```
pvj, line 33: I would try to word this differently. Leave empty to auto-increment or some such.

pvj (changes requested):
Thanks for this, it's good code! I am still bit unsure if this is the right place for this indeed needed functionality and would advocate for something else. But since it is small change I think we could start w/ it living here and continue discussions further. But even, before we merge it I would suggest we use queries.

gammaw:
I'm closing this, see details in https://github.com/mmmnmnm/lahmacun_arcsi/issues/14

Issue #25, opened 2020-10-22
Server error on listen/download (irregular)


gammaw:
@pvj I guess there was no issue for this yet, or was there? Is only listen or download too?

gammaw:
@gammaw yeah it happens for both use cases. 

Just tried for `/item/61/listen` and the result:

![image](https://user-images.githubusercontent.com/28865986/96975883-66ec6200-151b-11eb-9d1c-502df2c24870.png)

And for `/item/61/download` and the result was this: 

![image](https://user-images.githubusercontent.com/28865986/96975954-83889a00-151b-11eb-8a12-7db0a485931d.png)

gammaw:
@pvj I'm sure it's no surprise but here is the root cause so that we have it documented. 

This is the log from `/app/aguni.log` in the arcsi container: 

```
  File "/app/arcsi/api/item.py", line 167, in listen_play_file
    return send_file(
  File "/usr/local/lib/python3.8/site-packages/flask/helpers.py", line 635, in send_file
    fsize = file.getbuffer().nbytes
MemoryError
```

I'm sure that you're ahead of me debugging this. I'll try to help as this seems to be the main blocker of exposing arcsi on the frontend.

gammaw:
Does this help? https://stackoverflow.com/questions/24318084/flask-make-response-with-large-files

gammaw:
Above commit proposes a simple workaround for **download** via redirecting to the generated presigned DO URL. Note that the href can still be the API URL so that it remains persistent (always generates a fresh DO URL), looks nice (in the arcsi domain) and transportable (abstracts DO away). The expert user can see the redirect URL in the response but it shouldn't be a big deal. The pros are obviously that we don't have to download the file inside the app and the storage takes care of parallel users. 

For the **listen** part I could imagine an API which returns the a fresh DO URL and the client can embed it in an audio source HTML. An easy solution can be to do it from PHP but then the URL may expire. Alternatively, Vue (Javascript) can keep the URL fresh.

@pvj what do you think?

gammaw:
> For the listen part I could imagine an API which returns the a fresh DO URL and the client can embed it in an audio source HTML. An easy solution can be to do it from PHP but then the URL may expire. Alternatively, Vue (Javascript) can keep the URL fresh.

Added a small demonstrator for the easy solution via server code (here Flask but could be PHP in Wordpress).

gammaw:
[...]

PR #37, opened 2020-11-18
#25 server send file memory error

gammaw:


pvj (commented):
Thanks for the changes. Good solution! 

The PR contains changes that I believe we should discuss and perhaps consider to branch it out from here. Just pragmatic thought if we want to move fast w/ releasing it on the public front-end that shouldn't need to wait until we figure out answer to the following question:

Why did you add `file_url`? Could we maybe use the new solution you wrote for `download_play_file` method on the `listen_play_file` endpoint too, and remove the need for `file_url`? Or if it turns out to be needed then it should make all the way to the db layer I think so we can store it there and add date check to it. 

Both would work in a way where we don't have to generate it on every `view_item` call which is the direction I think we should move towards w/ all assets, audio & images once everything is settled. 

So the download changes are good how about we move forward w/ that? Or would you like to merge both download and listen at once?

gammaw:
> Why did you add file_url? Could we maybe use the new solution you wrote for download_play_file method on the listen_play_file endpoint too, and remove the need for file_url? Or if it turns out to be needed then it should make all the way to the db layer I think so we can store it there and add date check to it.

Good questions. The simple redirect works for download but I don't know if it would work for the listen use case (my understanding was that we need a URL for the listen embed). Yes, I guess file_url would work for download too but why make things complicated. I was also thinking of having file_url in the db but since it's a temporary URL I kept it in the JSON response only.

gammaw:
@pvj please review it again after the latest listen API fix.

pvj (approved):
Looks good to me~~

Issue #34, opened 2020-11-15
Add 1-day archive delay for listing


gammaw:
As we discussed, we filter the show archives based on their upload date and return them if day are > 1 day old.

PR #38, opened 2020-11-19
Issue #34 -- Quick workaround that filters future episodes 

pvj:
Still have not found time for reworking this queriless situation. 

However w/ list comprehension on the list of shows returning from JSON we can eliminate all episodes w/ `play_date` equalling `today()` or `play_date` set to the future

### DIFF
--- a/arcsi/api/show.py
+++ b/arcsi/api/show.py
@@ -3,7 +3,7 @@ import os
 import requests
 import io
 
-from datetime import datetime
+from datetime import datetime, timedelta
 from flask import flash, jsonify, make_response, request, url_for
 from flask import current_app as app
 from marshmallow import fields, post_load, Schema, ValidationError
@@ -96,7 +96,7 @@ def view_archive(slug):
     show = show_query.first_or_404()
     if show:
         show_json = show_details_schema.dump(show)
-        show_items = [show_item for show_item in show_json["items"] if datetime.strptime(show_item.get("play_date"), "%Y-%m-%d") < datetime.today()]
+        show_items = [show_item for show_item in show_json["items"] if datetime.strptime(show_item.get("play_date"), "%Y-%m-%d")+timedelta(days=1) < datetime.today()]
         for show_item in show_items:
             show_item["image_url"] = do.download(show.archive_lahmastore_base_url, show_item["image_url"])
         return json.dumps(show_items)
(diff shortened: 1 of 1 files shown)


gammaw:
@pvj this solution seems not to filter out episodes that are uploaded today (see example). As the play date doesn't seem to consider the show's hour (start, end), with this solution we would archive shows that will be broadcast later the day of the upload. Or am I missing something? 

An unrelated (perhaps silly/beginners' question): the show slug seems to be generated from the playlist name, not from the the show name (see code). Doesn't the frontend rely on show slugs generated based on the show's name (this is how the archive API is called, no)? 

```
@arcsi.route("/show/<string:slug>/archive", methods=["GET"])
def view_archive(slug):
    do = DoArchive()
    show_query = Show.query.filter_by(archive_lahmastore_base_url=slug)
```


image

image

pvj:
Another good point, added timedelta of 1 day but only to the comparison. This way if episode is played at one day earlier it still is returned since it is inclusive due to midnight hours implicit, but if it is played the same day as archive is requested it will be shifted by one day thus it won't be returned.

This shouldn't have to do anything w/ upload date. It's play date vs. the date on the server

As for your question: 
The slug is what we receive in our request as query, and to make it easy to connect with this very specific thing the general architecture we apply the same text manipulation when creating our archive base url (since it needs text manipulation anyway) that is used by the frontend to create its slugs. Comparing slugs we receive to names we have would fail and we can't expect anything else then slugs to be passed in request queries reliably (special char escaping, non-ascii chars, different text manipulation rules)

gammaw:
The fix looks good! Tried it too. 

> As for your question:
> The slug is what we receive in our request as query, and to make it easy to connect with this very specific thing the general architecture we apply the same text manipulation when creating our archive base url (since it needs text manipulation anyway) that is used by the frontend to create its slugs. Comparing slugs we receive to names we have would fail and we can't expect anything else then slugs to be passed in request queries reliably (special char escaping, non-ascii chars, different text manipulation rules)
> 

Sure. I thought for a moment that arcsi generates the slug (lahmastore_base_url) based on the playlist name (I had a local case which made me think that). But now I see it's not the case (see example). All good. Thanks. 

image

gammaw (approved):

pvj:
Cool! I'm glad you managed to try it out! Would make good show too, just the title sentence slowed down to speed of slugs

PR #39, opened 2020-11-19
Issue #36 -- Archive filename set by naming schema on upload

pvj:
We previously managed to control the file's name when user would download. Internally we still used the raw filename's of the assets during upload & archival.

The aim of this PR is to once and for all get rid of the raw filenames and set during upload names that are neutral. 

The schema is very simple. `<show_name>-<ep_name>.<ext>`

### DIFF
--- a/arcsi/api/utils.py
+++ b/arcsi/api/utils.py
@@ -70,8 +70,8 @@ def broadcast_audio(audio_file_path, item, image_file_path):
 
 
 def process_audio(play_file, item, image_file_path):
-    play_file_name = normalise(play_file.filename)
-    if play_file_name != "":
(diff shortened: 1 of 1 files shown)


gammaw:
> The schema is very simple. <show_name>-<ep_name>.<ext>

Sounds good. 

I tried and it seemed that the extension was not attached (mp3). Did it work for you?

pvj:
Good catch! It must be part of it explicitly. Rsplit should make sure regardless of the filename we always only split the extension type

gammaw (approved):
Fix looks good! I tried it on one MP3 with nasty chars in the episode name and it worked fine. Thanks!

PR #49, opened 2020-12-24
More conservative numbers for gunicorn

pvj:
We have smallish available memory so the upkeep for these workers normally used is bit too high. 

Lowered it and checked that memory is indeed not used by the extra workers and threads


pvj:
Failed to mention that this can be tested on current prod (should be too!)

gammaw:
@pvj I'm fine with the approach (and the PR) and I further lowered the redundancy. If you're OK with it, go ahead with PR.

gammaw:
We discussed compromise with 2 workers for minimal concurrency.

PR #50, opened 2020-12-24
#33 Setup new architecture w/ CDN edge

pvj:
Added new possibility to add CDN edge endpoint URL to the config which is then used when downloading files (saving costs and bandwidth). 
Main points:
- For this to work we now set public read ACL on the uploaded files as well as constructing the download path w/ the CDN endpoint URL
- We introduced ARCHIVE_ENDPOINT (perhaps not the best naming idea) as public URL media handler should use
- refactored storage download handler so it uses CDN (should revisit to add some fallback & smarts to this later)

This means once the already stored assets are set to public this can be merged and it can start receiving live requests I think

### DIFF
--- a/arcsi/handler/upload.py
+++ b/arcsi/handler/upload.py
@@ -71,18 +71,12 @@ class DoArchive(object):
             aws_secret_access_key=self.config["secret_key"],
         )
 
-        self.presigned = cli.generate_presigned_url(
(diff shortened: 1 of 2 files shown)


pvj:
This can be tested on the dev env where it is currently checked out

gammaw (approved):
Nice, logical, compact changes! Thanks and happy xmas ;)

PR #52, opened 2020-12-27
- API item.py: set empty string to kill root cause of bug for future …

gammaw:
- API item.py: set empty string to kill root cause of bug for future entries (redundant with View item.py – read on)
- View item.py: tolerate error for legacy entries


gammaw:
We discussed that it's OK to merge.

PR #56, opened 2021-01-16
Arcsi stops error

gammaw:
- DB version dump (10 -> 12)
- Debug mode default 
- Simple error handling for process media in add and edit APIs
   - Based on Boolean outputs of functions
   - In case of error, API returns an error string
   - Detailed infos can be found in aguni.log

### DIFF
--- a/arcsi/handler/upload.py
+++ b/arcsi/handler/upload.py
@@ -160,6 +160,8 @@ class AzuraArchive(object):
                         remaining_byte -= chunk_byte
                         start += chunk_byte
                         app.logger.debug("chunk posted! {} bytes remaining".format(remaining_byte))
+                    else: 
+                        app.logger.debug("response error code: {}".format(req.status_code))
                 except requests.exceptions.RequestException as e:
                     # max 30 retries
                     if retry >= 30:
--- a/arcsi/handler/upload.py
+++ b/arcsi/handler/upload.py
@@ -161,7 +161,8 @@ class AzuraArchive(object):
                         start += chunk_byte
                         app.logger.debug("chunk posted! {} bytes remaining".format(remaining_byte))
                     else: 
-                        app.logger.debug("response error code: {}".format(req.status_code))
+                        app.logger.debug("response error code: {}, body: {}".format(req.status_code, req.text))
+                        break #exit while loop when error
                 except requests.exceptions.RequestException as e:
                     # max 30 retries
                     if retry >= 30:
--- a/arcsi/handler/upload.py
+++ b/arcsi/handler/upload.py
@@ -161,8 +161,8 @@ class AzuraArchive(object):
                         start += chunk_byte
                         app.logger.debug("chunk posted! {} bytes remaining".format(remaining_byte))
                     else: 
-                        app.logger.debug("response error code: {}, body: {}".format(req.status_code, req.text))
-                        break #exit while loop when error
+                        app.logger.debug("Azuracast response error code: {}, body: {}".format(req.status_code, req.text))
+                        return False #exit while loop when error
                 except requests.exceptions.RequestException as e:
                     # max 30 retries
                     if retry >= 30:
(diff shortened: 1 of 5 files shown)


entrypoint.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 python3 -m flask db upgrade
-gunicorn --worker-tmp-dir /dev/shm -w 2 --worker-class=gthread --threads 2 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level info "arcsi:create_app('../config.py')"
+gunicorn --worker-tmp-dir /dev/shm -w 2 --worker-class=gthread --threads 2 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level debug "arcsi:create_app('../config.py')"
```
pvj, line 3: We shouldn't commit `debug` level, or?

arcsi/handler/upload.py
```diff
@@ -228,10 +231,10 @@ def assign_playlist(self,):
                 headers=self.config["headers"],
                 json=payload,
             )
-            if r.ok:
+            if r.ok and not self.empty_playlist():
```
pvj, line 234: That's smart way to ensure the addition of the new media was successful

arcsi/handler/upload.py
```diff
@@ -160,6 +160,9 @@ def upload(self,):
                         remaining_byte -= chunk_byte
                         start += chunk_byte
                         app.logger.debug("chunk posted! {} bytes remaining".format(remaining_byte))
+                    else: 
+                        app.logger.debug("Azuracast response error code: {}, body: {}".format(req.status_code, req.text))
+                        return False #exit while loop when error
```
pvj, line 165: Do we expect here that after exit the requestexception is raised and new round of retry happens?

pvj (approved):
The error handling is clean & simple and seems to cover all cases. I'm sorry there are still so many places w/ boilerplate or almost identical methods so these changes can't be done in 1 location. 

I have couple of questions / comments -- minors, so all in all I approve of the changeset. Great jobs w/ this~~~~

entrypoint.sh
gammaw: Actually, I thought yes. The log isn't too verbose even in debug mode (i.e., I see no disk concerns) but they can help us understand tricky cases during production (sharp operation). I see most debug log information we have now as "verbose" rather than "development/debug". What do you think?

arcsi/handler/upload.py
gammaw: No. The finding here was exactly that if a file cannot be uploaded due to tricky file name, then there will be no request exception, only the request will not be "ok" (false value). A retry would make no sense it this case because Azura will deterministically reject the file. 

But the exception case is good to have, I think. So it's good you added it. I'm not sure I can think of an easy case where it helps, but we never know.

entrypoint.sh
pvj: I see. In that case I might want to rename most of the current debug level logs to info and still change this back. Just the notion in my mind running anything on debug on prod is hard to swallow. But this is mostly aesthetics so we can go ahead w/ this

arcsi/handler/upload.py
pvj: What's the Az response like? status code is 200 and there's json payload w/ the message request? Yeah there is no point in the retry mechanism for rejected uploads. That only is supposed to help w/ network connection is an issue. We can later remove it or rewrite this a bit to use the json payload to catch exceptions on upload like this. 

But let's go ahead w/ this now I think. Thanks for the details these are extremely useful to know

entrypoint.sh
gammaw: I agree. One thing which is not fully cosmetical is the built in gunicorn debug log entries, which are very informative during normal operation, in my opinion (e.g., "Closed connection" every time a route request is served). What I mean is that if we change the above log entries to info mode, and we switch to info mode on the prod server, then these default gunicorn entries would disappear.

arcsi/handler/upload.py
gammaw: > What's the Az response like? status code is 200 and there's json payload w/ the message request?

Like this: https://github.com/mmmnmnm/lahmacun_arcsi/issues/54#issuecomment-761234391

gammaw:
@pvj thanks for the detailed review and for the good discussion! We can follow up on those still open issues (we can open new issues, for example) but I will merge this one for now as per your approval.

PR #61, opened 2021-02-13
Have a single worker

gammaw:
#57
Rationale: multiple workers are more expensive than multiple threads. Workers are recommended in case of CPU heavy load (which we don't have) and also if multiple cores are available (which we don't have either).


pvj (approved):
This is very weird and I am very sorry -- I was under the impression I already approved once PR regarding worker changes

PR #67, opened 2021-02-27
Form filenames universally

pvj:
Previously we already managed to clean up special chars and accents in uniform manner but we failed to address the logic of making sure we are always saving to the binary data to correct extension. What we had was one stop solution for audio assets but we didn't really care of images and it just worked. But with filename normalisation changes it break images and this is an attempt to fix that and provide a concise function that forms filenames correctly based on its input regardless of what extension it is. 

We are currently running both dev and sadly both prod too to not have too many broken images we need to fix later on this branch so if 1 days uploads is any verification than we can verify it works. It also handled cases w/ multiple dots in the filename well.
`some show e19. 20210220.MP3`
` some_show-episode_title_e19.MP3`

### DIFF
--- a/arcsi/api/utils.py
+++ b/arcsi/api/utils.py
@@ -48,7 +48,7 @@ def form_filename(file_obj, item_obj):
     Get the extension from file sent to API.
     To get the extension we use rsplit w/ maxsplit=1 to make sure we always get the extension even if there is another dot in the filename.
     '''
-    ext = file_obj.filename.rsplit(DOT, 1)[1]
+    ext = file_obj.filename.rsplit(DOT, 1)[1].lower()
     norms_show_name = normalise(item_obj.shows[0].name)
     norms_ep_name = normalise(item_obj.name)
     norms_names = [norms_show_name, norms_ep_name]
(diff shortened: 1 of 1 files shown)


gammaw (approved):
Very nice consolidation of the code!
One optimisation I could think of is that we compute `norms_names` only once. Currently, we do the same job twice, once for the audio, once for the image. But that's a minor issue, we can move on now.

pvj:
I was thinking we would need to do it separately in some instances but now
that I think about it I can't think of instances when they would get
different input. Feel free to add a small function that computes it but
please add it to both, since during live shows we might have an episode
image that we have to form while not having audio assets on those occasions

gammaw:
Yeah that's true with the live "empty" episodes. I think it's fine how it is now. It was just klugscheisserei :)

pvj:
Not at all. This was great point to make, its good to have review in code
review after all. Since we might add to this some more sanitisation before
we merge it, I think I'll add this one too!

pvj:
@gammaw There were so many issues coming out of the woodwork that we want to address with the asset management and media related refactor anyway that I couldn't find quick & clean way to fix your pain points. However I hope you may still accept this PR as is and additionally I added a small commit to make sure we will always create filenames with low cap extensions -- this way there won't be any issue with our Az integration anymore. Just tested it on the dev arcsi and broadcast servers successfully

gammaw:
I happily approve :) Looking forward to this version!

PR #80, opened 2021-04-04
#57 arcsi hangs again

gammaw:
Like we discussed, it's a temp fix to get rid off the triple-query on home. The html template has been initialised and the in-progress state is shown.


gammaw:
Like we discussed, it's a temp fix to get rid off the triple-query on home. The html template has been initialised and the in-progress state is shown.

PR #92, opened 2021-05-11
#89 usabily fixy

baxgas:
For now just small adjustments, but we can continue with the improvements, there is a lot of things to make it more user friendly


pvj:
@baxgas Thanks! Please just delete the code instead of commeting out. Afterwards if you feel okay about it you should merge it and we can deploy to prod!

PR #93, opened 2021-05-18
deleted commands refering to missing entrypoint.sh

tuz666:
Related: #86


pvj (approved):
Thanks! Long overdue!

PR #94, opened 2021-05-18
update to help future Windows developers

tuz666:
Related: #86
the .gitattributes file helps to evade Windows' auto CRLF line endings nightmare


pvj (approved):
Maybe some globs are still missing -- we can improve that later. 

For now w/ this setup and `autocrlf = false` added to `.gitconfig` `[core]` works for our sole Windows-based developer. 😉 

So I'd say it is safe to say this improves their lives and we should merge it.

gammaw:
Nice!

PR #98, opened 2021-05-21
API returns show episodes in descending order

pvj:
- Added utility method that helps us sort through list of dicts by value
- Used it on show view (only for now -- I think!) 
- No need for change on frontend (I hope!)


gammaw (approved):
Looks good! Tried on dev (both arcsi & frontend) and does the job. Looking forward to using it!

Issue #13, opened 2020-09-19
Add validation for audio files and images


gammaw:
By mistake, I uploaded an mp3 as cover image and the image as audio file. On DO side the result is correct (DO doesn't know what is what) but in Azuracast it results in a wrong state (the entry is there, the playlist is set, but an image is used for the audio file). 

Can we add some basic validation on the client side such that we check for **mp3 and mp4** for the audio file, and **jpeg and png** for the image? Actually, it would not be bad to disallow wav and other formats.

gammaw:
See related issue: https://github.com/mmmnmnm/lahmacun_arcsi/issues/18

tuz666:
We double checked it and it returns an error in case you switch the audio and image file, the form also suggests the recommended file types.

PR #99, opened 2021-05-21
Introducing extension validation 

pvj:
Currently we only check that file in request sent to API is present (ie. not empty)

I introduced a utility method that takes an allowed extension list (I put this in config for ease of working with it in the future, note: **after merge config update is needed on live instances**) and compares the filename extension to it

To solve the issue presented in #13 I created list for valid audio then list for valid image extensions and added checks to show and episode API's separately.

I think ideally this validation could/should happen right in the client but it's not too expensive and bad to have something defensive like this deep down in the API (we can always remove it later too). So I added it to our backend for now.

Another note. I could have updated the "error message" to reflect that we might now fail b/c of this but I'd rather focus on adding proper extension handling. 

Please do run 1 test or 2. Looked okay on my end but who knows.

### DIFF
--- a/arcsi/api/utils.py
+++ b/arcsi/api/utils.py
@@ -12,6 +12,13 @@ DELIMITER = "-"
 DOT = "."
 
 
+def allowed_file(filename):
+    return (
+        DOT in filename
+        and filename.rsplit(DOT, 1)[1].lower() in app.config["ALLOWED_EXTENSIONS"]
+    )
+
+
 def dict_to_obj(dict_name, table):
     show_seq = (k["id"] for k in dict_name)
     obj_list = db.session.query(table).filter(table.id.in_(show_seq)).all()
@@ -34,6 +41,7 @@ def normalise(namestring):
     norms = slugged.replace(DELIMITER, CONNECTER)
     return norms
 
+
 def slug(namestring):
     slugs = slugify(namestring)
     return slugs
@@ -54,17 +62,19 @@ def form_filename(file_obj, title_tuple):
     return "{}{}{}".format(DELIMITER.join(norms_names), DOT, ext)
 
 
-def broadcast_audio(archive_base, archive_idx, broadcast_file_name, broadcast_playlist, broadcast_show, broadcast_title, image_file_name):
+def broadcast_audio(
+    archive_base,
+    archive_idx,
+    broadcast_file_name,
+    broadcast_playlist,
+    broadcast_show,
+    broadcast_title,
+    image_file_name,
+):
     broadcast_file_path = media_path(
-        archive_base,
-        str(archive_idx),
-        broadcast_file_name
+        archive_base, str(archive_idx), broadcast_file_name
     )
-    image_file_path = media_path(
-        archive_base,
-        str(archive_idx),
-        image_file_name
-    )
+    image_file_path = media_path(archive_base, str(archive_idx), image_file_name)
     az = AzuraArchive(
         broadcast_file_path,
         broadcast_file_name,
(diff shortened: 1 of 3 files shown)


arcsi/api/utils.py
```diff
@@ -89,23 +99,24 @@ def broadcast_audio(archive_base, archive_idx, broadcast_file_name, broadcast_pl
...
-    archive_file_name = form_filename(archive_file, archive_name)
-    if archive_file_name != "":
-        archive_file_path = media_path(
-            archive_base, str(archive_idx), archive_file_name
-        )
-        archive_file.save(archive_file_path)
-        return archive_file_name
-    else:
+    if not allowed_file(archive_file):
         return None
```
gammaw, line 103: I was wondering where the corresponding error handling was. 

Here? 

https://github.com/mmmnmnm/lahmacun_arcsi/blob/024ba5ff1ecd0e773c4c5dde3068b09cc6b7e9dc/arcsi/api/item.py#L342-L349

Will a `None` return value be a Boolean `False` here? Btw, shouldn't it be `if not (image_file_name and not item.play_file_name)` (because we need both audio and image for a valid request)?

config.template.py
```diff
@@ -23,6 +21,7 @@
 
 # ARCSI CONF
 UPLOAD_FOLDER = "/abs/path/on/your/machine"  # this is where arcsi stores uploadable files temporarily
+ALLOWED_EXTENSIONS = ["mp3", "gif", "jpg", "jpeg", "png"]
```
gammaw, line 24: Isn't it strange to mix audio and image formats? Two lists (one for audio, one for image) would be more fault tolerant but I see that the current process() function doesn't know what it gets (image or audio), right?

gammaw (commented):
Added a couple of questions. 

I agree, a precise error message (which assumes a better error handling) would be nice to have. We're now adding another potential source of error without changing the generic error message. But at least we don't upload obvious bullshit.

config.template.py
pvj: It is indeed strange and I first wrote `AUDIO` and `IMG` lists but as you say that was bit too ahead of its time. We don't yet know what it gets -- we will (have to) though, when we are at a point where we want to optimise processing based on what we are dealing with. 

But for now I went w/ the simplest working iteration

arcsi/api/utils.py
pvj: The line you linked is the correct one, yes. `None` is `False` in Python in a boolean evaluation, yes. 

Yes we need both for valid requests but conversely I think that means that if either one of them, image file or play file couldn't be processed then we assume something went wrong and we shouldn't proceed. We don't need both to error out, one is enough, I think. Did I make a logical mistake?

This is of course quite rudimentary proper error propagation & handling should be a focus in the next release

gammaw: > The line you linked is the correct one, yes. None is False in Python in a boolean evaluation, yes.

Cool!

> Yes we need both for valid requests but conversely I think that means that if either one of them, image file or play file couldn't be processed then we assume something went wrong and we shouldn't proceed. We don't need both to error out, one is enough, I think. Did I make a logical mistake?

Yes, we need to return an error if either `image_file_name` or `item.play_file_name` is `None` (`False`). But in oder to satisfy `not (image_file_name or item.play_file_name)` you need both to be `False`.  If we express it with De Morgan's law, we need here `(not A) or (not B)` as the condition of the error which is equivalent with `not (A and B)`. But maybe I'm wrong again :)

pvj: No you are correct! Thanks~~~

PR #100, opened 2021-05-22
Fixing bad behaviour of edit show api

pvj:
**What was buggy?**
The endpoint wrongly used the value of the existing item's `archive_lahmastore` column to turn `archived` flag `True`. You might have just updated the image of a live show and it would still change like there was audio archived (thanks @tuz666 )

**What did I do?**
I just removed this incorrect statement. No idea why it was even there. I'm sorry for that.

I tested it locally w/ dev: 

Editing w/o audio now doesn't set it to archived
Editing w/ audio still does set it to archived correctly

Please let me know if you have more test cases and do try the branch out, cheers!

### DIFF
--- a/arcsi/api/item.py
+++ b/arcsi/api/item.py
@@ -177,7 +177,7 @@ def add_item():
                         archive_file=play_file,
                         archive_name=(new_item.shows[0].name, new_item.name),
                     )
-            
+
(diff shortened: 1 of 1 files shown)


gammaw (approved):
Good dig! Read your comment, checked the code, all make sense. Also tested it on a local setup and it behaves as expected.

PR #111, opened 2021-07-08
Fix issue to reach path to live ep editing

pvj:
**What happened?**
I made an indentation type so if you wanted to edit an episode into a live type one it wasn't possible. Thank you @gammaw for noticing this in #110 

**What did I do?**
Fixed the statement so it is now in it's hopefully correct pos

### DIFF
--- a/arcsi/api/item.py
+++ b/arcsi/api/item.py
@@ -182,8 +182,8 @@ def add_item():
                 # we require both image and audio if broadcast (Azuracast) is set
                 if not (image_file_name and new_item.play_file_name):
                     no_error = False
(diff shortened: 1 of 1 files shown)


gammaw:
Thanks for the fix! I also overlooked it in the first review. Shouldn't we introduce a must-test policy for each PR (maybe except the prog PRs)?

gammaw (approved):

pvj:
Yeas we definitely should especially b/c this is still such a hot mess. I think there might be scenarios where you don't have an image but try to create new ep / edit existing ep is valid scenario.

I'll merge this now since it is definitely improvement over an improperly placed else statement but will keep monitoring

PR #115, opened 2021-08-11
Added the cleaned version of our Swagger_UI

tuz666:
Nearly finished the whole documentation of Arcsi's API calls.
It seems that everything is working now.
I find some properties in the request post bodies which I thought not necessary.
I flagged them with a "fölösleges" tag, please correct me if I'm wrong.

### DIFF
--- a/arcsi/__init__.py
+++ b/arcsi/__init__.py
@@ -4,6 +4,7 @@ import os
 from flask import Flask
 from flask_security import Security, SQLAlchemySessionUserDatastore
 from flask_migrate import Migrate
+from flask_swagger_ui import get_swaggerui_blueprint
 from sqlalchemy.exc import ProgrammingError
 
 from arcsi.model import db, item, role, show, user
@@ -63,5 +64,17 @@ def create_app(config_file):
 
     app.register_blueprint(api.arcsi)
     app.register_blueprint(view.router)
+
+    ### swagger specific ###
+    SWAGGER_URL = '/docs'
+    API_URL = '/static/docs.json'
+    SWAGGERUI_BLUEPRINT = get_swaggerui_blueprint(
+        SWAGGER_URL,
+        API_URL,
+        config={
+            'app_name': "Arcsi"
+        }
+    )
+    app.register_blueprint(SWAGGERUI_BLUEPRINT, url_prefix=SWAGGER_URL)
 
     return app
--- a/arcsi/__init__.py
+++ b/arcsi/__init__.py
@@ -66,8 +66,8 @@ def create_app(config_file):
     app.register_blueprint(view.router)
 
     ### swagger specific ###
-    SWAGGER_URL = '/docs'
-    API_URL = '/static/docs.json'
+    SWAGGER_URL = '/doc'
+    API_URL = '/static/doc.json'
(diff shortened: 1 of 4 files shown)


arcsi/__init__.py
```diff
@@ -64,4 +65,16 @@ def create_app(config_file):
     app.register_blueprint(api.arcsi)
     app.register_blueprint(view.router)
 
+    ### swagger specific ###
+    SWAGGER_URL = '/docs'
```
pvj, line 69: I know this is a nitpick but can you please modify it from `docs` to `doc` ? We try to use an API endpoint naming schema where everything is singular

arcsi/static/docs.json
```diff
@@ -0,0 +1,960 @@
+{
+    "openapi": "3.0.0",
+    "info": {
+      "description": "Arcsi API doc",
+      "version": "0.9.0",
```
pvj, line 5: 😸

arcsi/static/docs.json
```diff
@@ -0,0 +1,960 @@
...
+              "type": "boolean"
+            },
+            "uploader": {
+              "type": "string"
+            },
+            "shows": {
+              "type": "integer",
+              "format": "int32"
+            },
+            "show_name": {
```
pvj, line 946: Where did you find this? I think you are right we can remove this

arcsi/static/docs.json
```diff
@@ -0,0 +1,960 @@
...
+              "type": "boolean"
+            },
+            "archive_mixcloud": {
+              "type": "boolean"
+            },
+            "users": {
+              "type": "integer",
+              "format": "int32"
+            },
+            "user_name": {
```
pvj, line 802: These are actually POST'ed to our API from the form, maybe put something of an example to these two? 

F.e.: `user_name` might be `Arcsi` and `user_email` `arcsi[at]lahmacun[dot]hu` ?

pvj (changes requested):
Hey tuz! Sorry for keeping you waiting this long.

Very very good job, wouldn't know if this is your 1st or 1000th pull request! I have 2 minor requests what do you think of those? Cheers

arcsi/static/docs.json
tuz666: I found them in the POST methods of the Show and Item DetailsSchemas
(arcsi/api/show.py line: 146-154, 229-237; arcsi/api/item.py line: 107-110, 290-293)
I tried to remove them but as I saw without them the methods won't work correctly.

arcsi/static/docs.json
tuz666: What's so funny here? 😨

arcsi/__init__.py
tuz666: Fixed it 👍

arcsi/static/docs.json
tuz666: I added every kind of examples to the PostRequestBodies

pvj (changes requested):
We did personal review of most of these and discussed how to continue please apply those as discussed. Thanks, great work!

pvj (approved):
@tuz666 You're really good librarian thanks for all the work on this! Mad!!

PR #121, opened 2021-09-25
Remove unique key constraint for playlist name in shows table

gammaw:
@pvj please review it. I could now assign the rare shows playlist to one new show on Thur and Fri and to the others. 

Edit: I didn't try it but I assume it would work.


gammaw:
Maybe I was being too naive here and just restarting the app won't change the schema in the running DB instance. Would it be one of those Alchemy migration steps needed here to update the DB? What do you think @pvj? I can go on trying it on dev if we're not sure enough.

gammaw:
@pvj thanks for the guidance (I created a quick Wiki based on it, see [here](https://github.com/mmmnmnm/lahmacun_arcsi/wiki/Apply-schema-change-to-DB)), I added a new migration version and tested successfully in a local instance. Please review.

pvj (approved):
If the marshmallow serialisation schemas are also not enforcing / expecting unique `playlist_name` then let's go. Sorry for the late review

EDIT: Also, great to the point wiki, thank you for that.

Issue #123, opened 2021-10-12
Improve arcsi rest endpoints


baxgas:
Just checked the swagger, nice job @tuz666 !
Would be nice to have a `shows/all?=basic` (or some other name) that would give back just the basic shows data (`name, description, time...`), without all the items (arcsi episodes). That would make the basic request for the ananasz calls much smaller.
For `item` (episode) would it be possible to have a queryable `slug` endpoint, so they can be queried also by this, not only the `id`? Also, if possible to have a `listening_count` besides `download_count` in response > don't know how it works in either case, how it counts this... At this query, it would be nice to have the `archive_lahmastore_base_url` at the `shows` array entries.

pvj:
Sorry for not responding here @baxgas 

Good suggestions. We are already looking into the first, trying to work out the cleanest solution. We might simply just not add the episodes to shows or list only their ids, instead of the full-blown object. We'll see and let you know.

Second suggestion is something we already do, no? If you check the items under a show, you get the same results as querying slug on items. Maybe I don't correctly understand the question. Sorry but please elaborate in that case.

>At this query, it would be nice to have the archive_lahmastore_base_url at the shows array entries.
Sorry, I might not understand this correctly. Do you want each show object to contain a self-reference?

We will be adding all these you asked or something very similar. 

If you can just help me bit with explanation that would be splendid. Thanks so much~~

pvj:
@baxgas please take a look at @tuz666 's nice PR and let us know what you think from the front-end side please

baxgas:
Thanks @pvj

gammaw:
Great efforts!

Can you help me to understand a couple of things. 

@baxgas writes:
> For item (episode) would it be possible to have a queryable slug endpoint, so they can be queried also by this, not only the id?

Why is this needed (query by slug vs query by ID)?  

@tuz666 writes (in the PR):
> added a new route /show_slug/episode_number where we can query the specified item directly

Don't you mean `/show_slug/episode_slug`?

pvj:
>Also, if possible to have a listening_count besides download_count in response > don't know how it works in either case, how it counts this... At this query, it would be nice to have the archive_lahmastore_base_url at the shows array entries.

I'll move this to a separate ticket because it's schema-related update to add `listening_count` 

Otherwise we merged this PR #127 just now so all others requests are done:

For the `show/all` we now only return: 

`"id", "active", "name", "description",
                                                         "week", "day", "start", "end",
[...]

PR #127, opened 2021-11-14
Fix/api endpoints

tuz666:
#123 Improve arcsi rest endpoints
Minor improvements in arcsi rest endpoints

- `show/all` route now sends back the `shows` data without the `items` field (the sixth of the size before)
- added a new route `/show_slug/episode_number` where we can query the specified item directly
- I think we should modify our db a little bit, imo there should be a unique field in the `Item` model as well. Maybe that would help us against the duplicated records

### DIFF
--- a/arcsi/api/__init__.py
+++ b/arcsi/api/__init__.py
@@ -9,3 +9,29 @@ from .user import *
 
 # TODO add routing here eg
 # arcsi.route("/archive", methods=["GET"])(<method_name>)
+
+
+@arcsi.route("show/<string:show_slug>/archive", methods=["GET"])
+def view_show_archive(show_slug):
+    show_query = Show.query.filter_by(archive_lahmastore_base_url=show_slug)
+    show = show_query.first_or_404()
+    if show:
+        # filter_date = datetime.today() - timedelta(days=1)
+        # airing_date = Item.play_date
+        # filter_hour = datetime.now() - timedelta(days=1)
+        # airing_hour = show.start
+        show_items = show.items.filter(Item.play_date + show.start < datetime.now() - timedelta(days=1)).all()
+        return jsonify(many_item_details_schema.dumps(show_items))
+    else:
+        return make_response("Show not found", 404, headers)
+
+
+@arcsi.route("show/<string:show_slug>/episode/<string:episode_slug>", methods=["GET"])
+def view_episode_archive(show_slug, episode_slug):
+    episode_slug += ".mp3"
+    item_query = Show.query.filter_by(archive_lahmastore_base_url=show_slug).join(Item, Show.items).first().items.filter_by(play_file_name=episode_slug)
+    item = item_query.first_or_404()
+    if item:
+        return jsonify(item_details_schema.dump(item))
+    else:
+        return make_response("Item not found", 404, headers)
--- a/arcsi/api/__init__.py
+++ b/arcsi/api/__init__.py
@@ -16,11 +16,7 @@ def view_show_archive(show_slug):
     show_query = Show.query.filter_by(archive_lahmastore_base_url=show_slug)
     show = show_query.first_or_404()
     if show:
-        # filter_date = datetime.today() - timedelta(days=1)
-        # airing_date = Item.play_date
-        # filter_hour = datetime.now() - timedelta(days=1)
-        # airing_hour = show.start
-        show_items = show.items.filter(Item.play_date + show.start < datetime.now() - timedelta(days=1)).all()
+        show_items = show.items.filter(Item.play_date < datetime.today() - timedelta(days=1)).all()
         return jsonify(many_item_details_schema.dumps(show_items))
     else:
         return make_response("Show not found", 404, headers)
--- a/arcsi/api/__init__.py
+++ b/arcsi/api/__init__.py
@@ -9,25 +9,3 @@ from .user import *
 
 # TODO add routing here eg
 # arcsi.route("/archive", methods=["GET"])(<method_name>)
-
-
-@arcsi.route("show/<string:show_slug>/archive", methods=["GET"])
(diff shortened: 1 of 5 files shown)


arcsi/api/item.py
```diff
@@ -402,3 +404,15 @@ def edit_item(id):
...
+
+
+@arcsi.route("/<string:show_slug>/<episode_number>", methods=["GET"])
+def view_episode_archive(show_slug, episode_number):
+    do = DoArchive()
+    # TODO instead of json filtering,
+    # write actual query
+    # joining shows and items
+    # so we can limit date etc.
+    item_query = Show.query.filter_by(archive_lahmastore_base_url=show_slug).join(Item, Show.items).filter_by(number=episode_number).first().items
```
pvj, line 414: Do we need `.items` objs here for the query execution below? Nice join!

arcsi/api/show.py
```diff
@@ -108,7 +108,7 @@ def view_show(id):
 
 
 @arcsi.route("/show/<string:slug>/archive", methods=["GET"])
-def view_archive(slug):
+def view_show_archive(slug):
```
pvj, line 110: Good naming schema thanks!

docker-compose.yml
```diff
@@ -33,7 +33,7 @@ services:
         - app.env
         - db.env
     # there is no `depend_on: condition:` in compose service v3
-    entrypoint: sh -c 'sleep 5; /app/entrypoint.sh'
+    entrypoint: sh -c 'sleep 60; /app/entrypoint.sh'
```
pvj: Yeah I'm sorry for this quite ugly hack with bit more work where the app just waits for db socket to be available this wouldn't be needed anymore.

pvj (approved):
Good changes addresses most of the issues from #123 very good work dude! Thanks so much.

One general nitpick / comment: 
I liked your PR description a lot however I felt it could be improved with additional test/example responses. Next time if you have bandwidth could you please try adding some? Thanks! 

Regarding your comment about Items. Valid point, I do see that we currently can't really use a field for querying unique items. But for that I don't think we need new unique columns as the id is already unique. I think we need a query logic on the API before the db layer which takes a request and gets couple of various values we can combine to make sure only 1 item ever is found as the result of the combination of these columns. We could tackle this in a new issue / branch next. What do you think?

arcsi/api/item.py
tuz666: Yes, without the `.first()` it returns with a `BaseQuery`, and without the `.items` it returns a `Show` object

arcsi/api/show.py
tuz666: Thanks, I tried to be as clear as I can. What do you think about rename this route to `"/<string:show_slug>/archive"` in order to abandon the redundant `/show` prefix?

pvj: There's a discussion in chat that I will join and we can discuss there

arcsi/api/item.py
pvj: This is a complex statement so of course there are multiple approaches but what about neither? If you just have the query that you then execute separately don't we save a transaction?

pvj (changes requested):
Thanks! Great job I think each change is in the right direction 

I added some comments and sorry about this but I have a change request: please make sure the archive date-limit stays in the logic of that endpoint. Let me know if it isn't clear!

tuz666:
> Good consolidation! With this change it will be lot more concise.
> 
> With the refactor it seems to me that the date-time based limit (expose archive episode only +1 day after airing) is missing.
> 
> Could you re-add it and validate that the output is the same?

I added this functionality with a little improvement, now it's checking the actual schedule of the show and filters according to that.
I couldn't test it already, but as soon as I can, gonna take a look on this.

gammaw:
> I added this functionality with a little improvement, now it's checking the actual schedule of the show and filters according to that.
> I couldn't test it already, but as soon as I can, gonna take a look on this.
> 

As discussed with @tuz666 too, we don't want to change the functionality now, so we can stick to the rule that a new episode will be available via arcsi starting **next midnight relative to play_date**. And only if the audio is available (it's another thing that we could improve upon later on, e.g., to list episodes where we deliverately don't want to share the audio).

arcsi/api/__init__.py
```diff
@@ -9,3 +9,25 @@
...
+        show_items = show.items.filter(Item.play_date < datetime.today() - timedelta(days=1)).all()
+        return jsonify(many_item_details_schema.dumps(show_items))
+    else:
+        return make_response("Show not found", 404, headers)
+
+
+@arcsi.route("show/<string:show_slug>/episode/<string:episode_slug>", methods=["GET"])
+def view_episode_archive(show_slug, episode_slug):
+    episode_slug += ".mp3"
+    item_query = Show.query.filter_by(archive_lahmastore_base_url=show_slug).join(Item, Show.items).first().items.filter_by(play_file_name=episode_slug)
```
pvj, line 28: I think we could simplify this by using the associative table for getting the `item_query` we want to execute. 

Something like `Item.query.join(items_shows).join(Show)`

And so this could be all joined together first and do the filtering on it after. Right now it is filter-join-filter.

But this might not work since we currently use `dynamic` setting for `lazy` relationship type instead of `joined` I'm sure though that we could get back an `item_query` without having to join `Item` to `Show.items` directly

arcsi/view/item.py
```diff
@@ -47,6 +47,7 @@ def view_item(id):
 
 
 @router.route("/item/<id>/edit", methods=["GET"])
+@roles_accepted("admin", "host", "guest")
```
pvj, line 50: Very very nice catch

pvj (approved):
After all the discussions I think these are some very good improvements that keeps stuff clean and simple. Let's move it in!

PR #131, opened 2021-12-05
Swagger update & minor fixes

tuz666:
I made some changes in Swagger:
- the new file structure separates the paths and the components making it easier to read
- it is updated to the new api endpoints

The `show/<show_slug>/archive` endpoint is fixed, now it returns a correct output


gammaw (approved):
Looks good to me! I tried it locally, the fixed endpoint now returns a JSON reponse, and I checked the Swagger page too. I see the point in decomposing the Swagger assets.

pvj:
Great job thanks for these!

Issue #128, opened 2021-11-26
For huge queries introduce pagination


pvj:
As the archive grows some endpoints became huge. It will only keep growing so it'd be best to limit the max items returned across the whole application.

This issue should be aimed at all list endpoints (items|shows|users|etc) 

In order for this feature to work we need minimum 1 parameter that lets us know the location of pagination. Like `?page=1` or `starts_at=123456789`

The argument we get for this parameter should determine which slice of all results we should return. 

I think one way that we could achieve it is the following: 
 - We should create a value for how many rows will be retrieved, like `list_limit` or similar (this will be used in our SQL statement as `LIMIT x`)
 - Another value would be keeping in store the current position, like `list_number` or something 
 - Making calculations with `list_limit*list_number` and assign it to some value which could be anything like `list_offset` or `list_position` or I don't know (this will be used to calculate SQL statement `OFFSET x`'s value)
   - These all should be of `Int` type
 - We create the query and add the calculated values, I don't remember the exact syntax but I'll write a couple that might have worked for me in the past, so e.g. `Sometable.query.all()[list_limit:list_offset]` || `Sometable.query.filter_by(something)[list_limit:list_offset]` || `Sometable.query.limit(list_limit).offset(list_offset).all()` or any of these not querying `Sometable` but our `db.session`

We may optionally provide a 2nd parameter that lets us specify something like `number=12` what can be used to inform us how many items should be returned.

For simplicity's sake I'd now only add 1 param and have a fixed size of returned data.

Let me please know if any info is missing or confusing! Go!

pvj:
This has been done marvelously well by @tuz666 so it goes without saying we could close

PR #132, opened 2021-12-12
Dev/paginate and more effective routes

tuz666:
Added pagination for the `/item/all` route #128 
Made a lot of optimizations in the API calls, to only return the neccessary data
Updated Swagger to make it more intuitive and readable for front-end developers

### DIFF
--- a/arcsi/api/show.py
+++ b/arcsi/api/show.py
@@ -15,7 +15,7 @@ from arcsi.handler.upload import DoArchive
 from arcsi.model import db
 from arcsi.model.show import Show
 from arcsi.model.user import User
-from arcsi.api.item import Item, many_item_details_schema, item_details_schema
+from arcsi.api.item import item_basic_details_schema
 
 
 class ShowDetailsSchema(Schema):
@@ -39,14 +39,12 @@ class ShowDetailsSchema(Schema):
         fields.Nested(
             "ItemDetailsSchema",
             only=(
-                "id",
-                "archived",
                 "description",
                 "name",
-                "number",
                 "play_file_name",
                 "play_date",
                 "image_url",
+                "download_count"
             ),
         ),
         dump_only=True,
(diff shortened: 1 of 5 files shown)


arcsi/api/show.py
```diff
@@ -85,28 +91,19 @@ def list_shows():
             show.cover_image_url = do.download(
                 show.archive_lahmastore_base_url, show.cover_image_url
             )
-    return many_show_details_basic_schema.dumps(shows)
+    return shows_schedule_schema.dumps(shows)
 
-@arcsi.route("/show/<id>", methods=["GET"])
-def view_show(id):
+
+@arcsi.route("/shows/archive", methods=["GET"])
```
pvj, line 97: I misunderstood our communication on chats a bit, I thought you replaced `/show/<id>` with `/show/<slug>` this way however `/shows/archive` doesn't quite replace `/show/<id>`

A simple way to see could be to check the function call instead of the route. The one we would like to replace is `view_show(id)` which gives us the data of a single show from the db (with its episodes included in a hacky wrong way). The one you've added is called `list_shows_archive()` and has two main differences: it is defined without the `id` parameter and is querying all of Show.

I don't mean by this that your work was for nothing and we revert this, I think a lightweight list of all shows you created is very useful for a response to some listing requests

arcsi/api/show.py
```diff
@@ -64,19 +62,27 @@ def make_show(self, data, **kwargs):
...
+                                                    "frequency", "language",
+                                                    "active", "description",
+                                                    "items"))
+show_partial_schema = ShowDetailsSchema(partial=True)
+shows_schema = ShowDetailsSchema(many=True)
+shows_schedule_schema = ShowDetailsSchema(many=True, 
+                                                   only=("active", "name", "cover_image_url",
+                                                         "day", "start", "end",
+                                                         "description", "archive_lahmastore_base_url"))
+shows_archive_schema = ShowDetailsSchema(many=True,
```
pvj, line 77: One small note here: Is this `archive` schema and the `/show/archive` endpoint for the front-end Shows list page? Maybe I misunderstand but in that case maybe we could rename it.

arcsi/api/show.py
```diff
@@ -262,27 +259,25 @@ def edit_show(id):
...
         )
 
-
 @arcsi.route("show/<string:show_slug>/archive", methods=["GET"])
 def view_show_archive(show_slug):
     show_query = Show.query.filter_by(archive_lahmastore_base_url=show_slug)
-    show = show_query.first_or_404()
+    show = show_query.first()
     if show:
-        show_items = show.items.filter(Item.play_date < datetime.today() - timedelta(days=1)).all()
```
pvj, line 274: Sorry if I missed out on this but is it okay to remove this date restriction?

pvj (changes requested):
Thanks for the all around improvement of code qual and delivery of this important feature!

There are some minor changes mainly on the API endpoints that I had to touch. There were 2 main reasons for that. 1) was that the changes removed actively used functionality (f. e. `view_show`) and 2) after the API change their call locations weren't updated (f.e. templates or view routers using `url_for(arcsi.view_show, id=id)`) to use the new routes.

The schema rework is awesome, and the new API endpoints as I said above should be very useful for the front-end. The icing is definitely the swagger update that should make life so easy from now on. Great work on all fronts!

pvj (approved):
I've added some compatibility fixes for now so we can merge and changed the shows endpoints to show. I might have to HOTFIX something as I didn't have time to test end-to-end but it is mostly just reverting stuff so I'm confident to approve

PR #133, opened 2021-12-13
Fix/Updated API endpoints and Swagger doc

tuz666:
In the previous PR, I forget about Arcsi's view methods & internal pages.
I restored the routes that were used by these features mentioned before.
The Swagger doc is more expressive than ever, it explains the exact use case and responses for every call.
Added the page size parameter for `/item/all_minimal/`
#128

### DIFF
--- a/arcsi/api/item.py
+++ b/arcsi/api/item.py
@@ -59,11 +59,11 @@ class ItemDetailsSchema(Schema):
 
 
 item_schema = ItemDetailsSchema()
-item_archive_schema = ItemDetailsSchema(only = ("name", "number", "play_date", "language", 
+item_minimal_schema = ItemDetailsSchema(only = ("name", "number", "play_date", "language", 
                                              "description", "image_url", "play_file_name", "download_count"))
 item_partial_schema = ItemDetailsSchema(partial=True,)
 items_schema = ItemDetailsSchema(many=True)
-items_basic_schema = ItemDetailsSchema(many=True, 
+items_minimal_schema = ItemDetailsSchema(many=True, 
                                                    only=("name", "description",
                                                          "play_date", "play_file_name",
                                                          "image_url", "download_count"))
@@ -75,15 +75,27 @@ headers = {"Content-Type": "application/json"}
 @arcsi.route("/item/all", methods=["GET"])
 def list_items():
     do = DoArchive()
+    items = Item.query.all()
+    for item in items:
+        if item.image_url:
+            item.image_url = do.download(
+                item.shows[0].archive_lahmastore_base_url, item.image_url
+            )
+    return items_schema.dumps(items)
+
+@arcsi.route("/item/all_minimal/", methods=["GET"])
+def list_items_minimal():
+    do = DoArchive()
     page = request.args.get('page', 1, type=int)
+    size = request.args.get('size', 12, type=int)
     items = Item.query.order_by(Item.play_date.desc()).paginate(
-        page, app.config['PAGE_SIZE'], False)
+        page, size, False)
     for item in items.items:
         if item.image_url:
             item.image_url = do.download(
                 item.shows[0].archive_lahmastore_base_url, item.image_url
             )
-    return items_basic_schema.dumps(items.items)
+    return items_minimal_schema.dumps(items.items)
 
 
 @arcsi.route("/item/<id>", methods=["GET"])
--- a/arcsi/api/item.py
+++ b/arcsi/api/item.py
@@ -59,11 +59,11 @@ class ItemDetailsSchema(Schema):
 
 
 item_schema = ItemDetailsSchema()
-item_minimal_schema = ItemDetailsSchema(only = ("name", "number", "play_date", "language", 
+item_archive_schema = ItemDetailsSchema(only = ("name", "number", "play_date", "language", 
                                              "description", "image_url", "play_file_name", "download_count"))
 item_partial_schema = ItemDetailsSchema(partial=True,)
 items_schema = ItemDetailsSchema(many=True)
-items_minimal_schema = ItemDetailsSchema(many=True, 
+items_archive_schema = ItemDetailsSchema(many=True, 
                                                    only=("name", "description",
                                                          "play_date", "play_file_name",
                                                          "image_url", "download_count"))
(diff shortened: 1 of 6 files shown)


arcsi/api/show.py
```diff
@@ -282,13 +297,45 @@ def view_show(id):
...
+# This will be the one that we are gonna use at the new page 
+@arcsi.route("show/<string:show_slug>/page", methods=["GET"])
+def view_show_page(show_slug):
     show_query = Show.query.filter_by(archive_lahmastore_base_url=show_slug)
     show = show_query.first()
     if show:
-        show_items = show.items.filter(Item.play_date < datetime.today() - timedelta(days=1)).all()
-        return items_basic_schema.dump(show)
+        # subquery = session.query(Item.id).filter(blabla -timedelta(day=1)).all().subquery()
+        # query = session.query(Show).filter_by(blabla).(Item.id.in_(subquery))
```
pvj, line 336: Would you like to finish this in this branch or in another later?

arcsi/static/component.json
```diff
@@ -189,7 +189,7 @@
                     }
                 }
             },
-            "ShowRequest": {
+            "ShowSubPageRequest": {
```
pvj, line 192: We used the docs today and it was very clean and easy to browse. Thanks a lot for building this library of knowledge!!!

config.template.py
```diff
@@ -45,5 +45,3 @@
 
 PORT = 5666  # The application port nginx proxy is passing to
 APP_BASE_URL = "http://{}:{}".format(HOST, PORT)
-
-PAGE_SIZE = 12
```
pvj, line 49: If there are no `args` in the URL does sqlalchemy also provide default value to `size`? Depending on that we don't need this as a fallback and we can remove it

arcsi/api/show.py
```diff
@@ -94,8 +109,8 @@ def list_shows():
     return shows_schedule_schema.dumps(shows)
 
 
-@arcsi.route("/show/archive", methods=["GET"])
-def list_shows_archive():
+@arcsi.route("/show/all_page", methods=["GET"])
```
pvj, line 112: Why not use this function body of this method as `/show` && `/show/all`'s body?

arcsi/api/show.py
```diff
@@ -282,13 +297,45 @@ def view_show(id):
...
+        return make_response("Show not found", 404, headers)
+    #show = show_query.first()
+    #if show:
+    #    show_items = show.items.filter(Item.play_date < datetime.today() - timedelta(days=1)).all()
+    #    return items_schema.dump(show_items)
+    #else:
+    #    return make_response("Show episodes not found", 404, headers)
+
+# This will be the one that we are gonna use at the new page 
+@arcsi.route("show/<string:show_slug>/page", methods=["GET"])
```
pvj, line 330: By the time we use the new frontend we should have `archive` url route free. But I think neither `page` nor `archive` are smooth sounding in this context but they were important that we started a discussion about this.

In my imagination these are rather named like `detail` or something. 
Also having a `/page?page=100` might be confusing

pvj (approved):

arcsi/api/item.py
```diff
@@ -74,16 +74,28 @@ def make_item(self, data, **kwargs):
...
+    do = DoArchive()
+    items = Item.query.all()
+    for item in items:
+        if item.image_url:
+            item.image_url = do.download(
+                item.shows[0].archive_lahmastore_base_url, item.image_url
+            )
+    return items_schema.dumps(items)
+
+@arcsi.route("/item/all_latest/", methods=["GET"])
```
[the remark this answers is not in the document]
tuz666: Yes, I only used `all` to represent that the response is not a single element

arcsi/api/item.py
```diff
@@ -74,16 +74,28 @@ def make_item(self, data, **kwargs):
...
+def list_items_latest():
     do = DoArchive()
     page = request.args.get('page', 1, type=int)
+    size = request.args.get('size', 12, type=int)
```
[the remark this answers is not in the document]
tuz666: @baxgas mentioned this as a nice to have feature

arcsi/api/show.py
```diff
@@ -62,7 +63,10 @@ def make_show(self, data, **kwargs):
         return Show(**data)
 
 
-show_schema = ShowDetailsSchema()
+show_schema = ShowDetailsSchema(only=("id", "name", "active", "description",
```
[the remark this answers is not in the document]
tuz666: Thanks for the information, it makes sense, gonna change this in the next commit

arcsi/api/show.py
```diff
@@ -84,6 +88,17 @@ def make_show(self, data, **kwargs):
...
+    do = DoArchive()
+    shows = Show.query.all()
+    for show in shows:
+        if show.cover_image_url:
+            show.cover_image_url = do.download(
+                show.archive_lahmastore_base_url, show.cover_image_url
+            )
+    return shows_schema.dumps(shows)
+
+@arcsi.route("/show/all_schedule", methods=["GET"])
```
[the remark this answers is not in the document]
tuz666: That's an interesting idea, but if we do this way then we will lose the response examples belonging to the different API calls in Swagger, what do you think about this?

arcsi/api/show.py
tuz666: Yes, I'll give it a try with the next commit.

arcsi/static/component.json
tuz666: My pleasure!

config.template.py
tuz666: Yes, exactly!

arcsi/api/show.py
tuz666: You are right, I defined a new `get_items()` method in the utils.py.

arcsi/api/show.py
tuz666: I got your point, and it's true again. I'm gonna replace the `page` to `subpages` this time, let's continue this discussion!

pvj (approved):

arcsi/api/show.py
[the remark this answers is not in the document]
gammaw: > We are basically slicing the same data, so most of the difference would be in schema applied. We could still control the flow with if-expressions. For example: /item/all?by=latest (diff. example: /item/all?by=2020-04-03) or /show/all?display=schedule (diff.: /show/all?display=list) etc. This is just a conversation starting idea let's not rush defining this well.

This would look more logical to me, I guess.

Issue #124, opened 2021-10-13
Add Tags option to Shows


baxgas:
store in array, with REST API endpoint access.
For now two entries: `Music`, `Talk` ÷ but with option to add freely any tag (`select` and `input` fields?)

gammaw:
@baxgas for now can we use WP to query the show tags?

baxgas:
@gammaw we don't use shows list from the wp page, only arcsi
in wp REST the posts are queried by ID, 
`$ curl https://example.com/wp-json/wp/v2/posts/<id>`
Have no idea how we could programmatically figure out this from arcsi shows. It is much easier to implement this in arcsi db, than to hack around and make cross-checks or whatever

tuz666:
Closed by @pvj as part of https://github.com/lahmacunradio/arcsi/pull/166!

PR #134, opened 2021-12-14
Tag feature for shows

pvj:
This PR is about adding functionality for tagging shows with a list of words, according to what we discussed in #124  

Tags are assigned to many shows. Shows can have more than one tag too. Later we can also easily extend this to episodes -- it is only a business decision, technically we are ready to go with both.

In order to model the many-to-many relation I created a new table called `tags` with class name `Tag`. The model has few attrs, the schema is simple too, and there is only a "demo" API call for separately demonstrating the tag collection however because of the schema and this we created the `tag` module in the apis. Last and least, I created a small model utility tool, so now there's a `model/utils.py` helper package for ORM & Db related things.

I migrated our schema and the db upgrade worked well on staging arcsi.

Feel free to test it there, and let me know if you have any questions.


arcsi/api/tag.py
```diff
@@ -0,0 +1,23 @@
...
+    
+    @post_load
+    def make_tag(self, data, **kwargs):
+        return Tag(**data)
+
+tags_details_schema = TagDetailsSchema()
+tags_partial_schema = TagDetailsSchema(partial=True)
+many_tags_schema = TagDetailsSchema(many=True)
+
+@arcsi.route("/tags", methods=["GET"])
```
gammaw, line 19: @pvj nice demo API! But why demo? Wouldn't we need it for example to list all shows with the same tag (a feature we have in the current website)? And for that wouldn't it be easier that the API returns the secondary relations?

gammaw (approved):
@pvj what a strong PR! Crisply documented, cleanly and logically designed and coded! Thanks a lot!

The only doubt I have is the richness of the new APIs wrt the "list all shows with the same tag" feature but that's a question we can discuss separately and enrich the APIs if needed. 

Scope of review: I locally tested the edit show use case, I tried the show and the demo tag APIs.

arcsi/api/item.py
```diff
@@ -59,6 +59,12 @@ class ItemDetailsSchema(Schema):
         ),
         required=True,
     )
+    tags = fields.List(
```
tuz666, line 62: Nice, we can use tags for episodes in the future as well. This will open a lot of new opportunities for the listeners to surf between episodes of different shows and give some interesting, new content for the show and episode subpages too.

arcsi/templates/show/add.html
```diff
@@ -44,6 +44,11 @@ <h2>{% block title %}Add radio show to archive{% endblock %}</h2>
       </select>
   </div>
 
+  <div>
+    <label for="taglist">Show tag list (separate with commas)</label>
```
tuz666, line 48: We should add these modifications for the episodes/items in the future.

arcsi/api/tag.py
pvj: @gammaw "Demo" in the sense that it just lists the existing tags from the database for proof of concept that the API works. 

Now I'll add the tag details page that will return those relations and then I'll consider it more than a demo

arcsi/model/utils.py
```diff
@@ -0,0 +1,10 @@
+from . import db
+
+def get_or_create(model, **kwargs):
```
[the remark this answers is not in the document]
pvj: @gammaw Yeah totally agreed with that, will start on this PR's new material

arcsi/templates/show/add.html
pvj: Sure, good idea why not have them there!

arcsi/api/show.py
```diff
@@ -61,6 +62,12 @@ class ShowDetailsSchema(Schema):
         ),
         required=True,
     )
+    tags = fields.List(
```
tuz666, line 65: It would be better to return the id of the tag here too and we could expand the schemas with the tag field as well (as I see it is enough to modify show_archive_schema since it is the one that supposed to be used for the show subpages).

arcsi/api/tag.py
```diff
@@ -0,0 +1,23 @@
+from marshmallow import fields, post_load, Schema
+
+from arcsi.api import arcsi
+from arcsi.model.tag import Tag
+
+
+class TagDetailsSchema(Schema):
+    id = fields.Int()
+    display_name = fields.Str(required=True)
```
tuz666, line 9: Do not forget about the empty string tag 😄

tuz666 (changes requested):

PR #135, opened 2021-12-14
Create new file name in case filename already exists

pvj:
In case that we accidentally send a file to the archives that has been already submitted we now change the filename so it can duplicate safely. 

I renamed one utility function `process` to `save_file` as that is way more exact.

The question is what to do on retrieval & display? 
In use-case one, the show host has a late edit (`mixdown-fin-2` etc.) he wants to upload. In that case the file exists so we change file name of the edited episode and upload it. In this case we would need to fetch this new copy

In another case, someone makes a mistake during the upload process and due to that we accidentally tried uploading something so again the name is safely changed but here we want the opposite to happen: retrieve & display the actual episode.

I'm curious what you think or can think of? I'll try to rest and think this through

### DIFF
--- a/arcsi/api/item.py
+++ b/arcsi/api/item.py
@@ -10,12 +10,14 @@ from flask import flash, jsonify, make_response, request, send_file, url_for, re
 from flask import current_app as app
 from marshmallow import fields, post_load, Schema, ValidationError
 
+from uuid import uuid4
+
 from .utils import (
     archive,
     broadcast_audio,
     dict_to_obj,
     media_path,
-    process,
+    save_file,
 )
 from arcsi.api import arcsi
 from arcsi.handler.upload import DoArchive
@@ -177,24 +179,37 @@ def add_item():
 
         # TODO get show cover img and set as fallback
         if request.files:
+            # Defend against possible duplicate files
+            name_occurrence = int(db.session.query(db.func.count()).filter(new_item.name == Item.name, new_item.number == Item.number).scalar())
+            version_pre = None
+
+            if name_occurrence:
+                version_prefix = uuid4()
+                item_name = "{}-{}".format(version_prefix, new_item.name)
+            else:
+                item_name = new_item.name
+
             # process files first
-            if request.files["image_file"]:
-                if request.files["image_file"] != "":
-                    image_file = request.files["image_file"]
-                    image_file_name = process(
-                        archive_base=new_item.shows[0].archive_lahmastore_base_url,
-                        archive_idx=new_item.number,
-                        archive_file=image_file,
-                        archive_name=(new_item.shows[0].name, new_item.name),
-                    )
             if request.files["play_file"]:
                 if request.files["play_file"] != "":
-                    play_file = request.files["play_file"]
-                    new_item.play_file_name = process(
+                    play_file = request.files["play_file"]  
+
+                    new_item.play_file_name = save_file(
                         archive_base=new_item.shows[0].archive_lahmastore_base_url,
                         archive_idx=new_item.number,
                         archive_file=play_file,
-                        archive_name=(new_item.shows[0].name, new_item.name),
+                        archive_name=(new_item.shows[0].name, item_name),
+                    )
+
+            if request.files["image_file"]:
+                if request.files["image_file"] != "":
+                    image_file = request.files["image_file"]
+
+                    image_file_name = save_file(
+                        archive_base=new_item.shows[0].archive_lahmastore_base_url,
+                        archive_idx=new_item.number,
+                        archive_file=image_file,
+                        archive_name=(new_item.shows[0].name, item_name),
                     )
 
             if new_item.broadcast:
@@ -205,6 +220,7 @@ def add_item():
             else: 
                 if not image_file_name:
                     no_error = False
+
         # archive files if asked
         if new_item.archive_lahmastore:
             if no_error and (play_file or image_file):
@@ -228,6 +244,7 @@ def add_item():
                         new_item.archived = True
                     else:  # Upload didn't succeed
                         no_error = False
+
         # broadcast episode if asked
         if new_item.broadcast and no_error:
             if not (play_file and image_file):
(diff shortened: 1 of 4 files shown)


arcsi/api/item.py
```diff
@@ -165,24 +167,37 @@ def add_item():
 
         # TODO get show cover img and set as fallback
         if request.files:
+            # Defend against possible duplicate files
```
gammaw, line 170: Isn't this check the same here and in edit? If yes, can you define it only once and use it where it is needed?

arcsi/api/utils.py
```diff
@@ -119,19 +119,19 @@ def broadcast_audio(
     return False
 
 
-def process(archive_base, archive_idx, archive_file, archive_name):
-    archive_file_name = form_filename(archive_file, archive_name)
+def save_file(archive_base, archive_idx, archive_file, archive_file_name):
```
gammaw, line 122: Does this code work for you? When I ran my first test upload it returned with an error on https://github.com/mmmnmnm/lahmacun_arcsi/blob/ead1b178daad65a1f05900e064754fe65620065f/arcsi/api/item.py#L189 because of TypeError. It's due to `archive_name` in the call vs `archive_file_name` in the signature, I guess, but maybe I'm wrong.

gammaw (commented):
@pvj thanks for the changes!  

I like your approach of keeping track of the file names locally in the DB without having to check in the cloud storage. It's correct as long as the DB is in synch with the storage but since the app is the only interface to the storage, it's a correct assumption. 

> In another case, someone makes a mistake during the upload process and due to that we accidentally tried uploading something so again the name is safely changed but here we want the opposite to happen: retrieve & display the actual episode.

I'm not sure I understood this use case. What mistake do you mean exactly? Do you mean that we create the duplicate in the DB (locally) but we fail to upload the file to the storage (e.g., due to network errors), which results in an out of synch state?

arcsi/api/utils.py
pvj: @gammaw Sorry for the letdown on first try. I'm trying to retrace my steps, clearly something went wrong. I tested it extensively locally and on staging env too -- there are actually couple of hashed duplicates in our storage & db. I don't know why these argument keywords were missing. Did I change the param name in a later commit? I must have, because certainly the local state of apis wasn't saved. One quick remedy is that I've turned on autosave now. On dev I must have tried the uploads on a previous commit. An improvement here I'll make is to add the commit to the footer. Will create an issue for this later. 

I'd say it was more than an educated guess, it was the correct one. You are exactly right and I think that's a great thing. Thanks for the great debugging! I pushed a supposed fix.

arcsi/api/item.py
pvj: Yes, that is long overdue. The reason I am hesitant to get at it or start a level-headed discussion because I want to do this by defining a separate media table put all this media server logic from the different object API's to their own. 

I would have added that "on the other hand -- it's OK not to touch it until this overhaul because it's trivial to update all the endpoints and apply the changes to their logic" but after tonight's miss on endpoint arguments I think it might be not that valid anymore and perhaps a good time to revisit this. 

Thanks for starting this train of thought!

arcsi/api/item.py
```diff
@@ -177,24 +179,37 @@ def add_item():
...
+            name_occurrence = int(db.session.query(db.func.count()).filter(new_item.name == Item.name, new_item.number == Item.number).scalar())
+            version_pre = None
+
+            if name_occurrence:
+                version_prefix = uuid4()
+                item_name = "{}-{}".format(version_prefix, new_item.name)
```
gammaw, line 188: @pvj how about putting the uuid4 in the end (as a postfix)? The advantage would be that if somebody downloads the file, the show and episode names are easily visible.

gammaw (commented):
@pvj we are getting there! After a local change I could test it in action, looks nice in general. I added some questions, can you check them and tell what you think?

tuz666:
It looks good to me but when I tried to test it on devarcsi something didn't work for me, I always got a DO error during the upload process. Can you give me a helping hand @pvj ?

arcsi/api/utils.py
```diff
@@ -119,19 +119,19 @@ def broadcast_audio(
...
+    formed_file_name = form_filename(archive_file, archive_file_name)
     if not allowed_file(archive_file_name):
```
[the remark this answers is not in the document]
gammaw: > @pvj did it work for you like this? Isn't it supposed to be formed_file_name which is checked against the allowed extensions? When I changed it worked for me.

Fixed here 5dd47fbe8770363fc3965e75e3821c4d7fcc37fb

arcsi/api/item.py
```diff
@@ -177,24 +179,37 @@ def add_item():
 
         # TODO get show cover img and set as fallback
         if request.files:
+            # Defend against possible duplicate files
+            name_occurrence = int(db.session.query(db.func.count()).filter(new_item.name == Item.name, new_item.number == Item.number).scalar())
+            version_pre = None
+
+            if name_occurrence:
```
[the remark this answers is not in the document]
gammaw: The reason seems to be that flush() adds the item and the query returns the result of the flushed transactions. See 03057739186dcf829ed00071cca3d7b1f4c795f3

arcsi/api/item.py
gammaw: Changed so that hash is appended as file name postfix: 8657aa46b803b54ae5152800e873a3960b44d081

gammaw:
After addressing some issues left in this one and after successful and independent tests by @tuz666 and me, @pvj @tuz666 @baxgas please review asap.

pvj (commented):
I can't approve my own PR but the changes look good @gammaw  thank you!

The only change I'd make but that can be done in the future would be to make `name_occurrence` a method of the `Item` class (or `utils` if we want to reuse it for shows or etc.) so we don't have to maintain the comparison query in two places.

tuz666 (approved):
I tested this branch on the dev, worked as planned. @pvj is right we should make some refactor here in the future.

gammaw:
Quick hands @tuz666 ;) I was going to make the proposed change but I guess I'll have to defer it. Merged branches cannot be merged again even with new (unmerged) commits and for this change I will not open a new branch.

gammaw:
> The only change I'd make but that can be done in the future would be to make name_occurrence a method of the Item class (or utils if we want to reuse it for shows or etc.) so we don't have to maintain the comparison query in two places.

Thanks @tuz666 for the [fix](https://github.com/mmmnmnm/lahmacun_arcsi/pull/141)!

PR #136, opened 2021-12-16
Hotfix for show api added function calls

tuz666:
initialized get_shows() functions in the api calls


pvj (approved):
I am sorry for this oversight in the PR process. Thanks for catching it  @tuz666

PR #137, opened 2021-12-17
Fixed Arcsi UI

tuz666:
During a previously merged PR I broke some navigation links in the Arcsi UI what we use to upload new shows & episodes and test new features.
Now I fixed the Show model schema: I re-added the nested id and number values to its items field, which were used by the view functions.

### DIFF
--- a/arcsi/api/show.py
+++ b/arcsi/api/show.py
@@ -40,7 +40,9 @@ class ShowDetailsSchema(Schema):
         fields.Nested(
             "ItemDetailsSchema",
             only=(
+                "id",
                 "description",
+                "number",
                 "name",
                 "play_file_name",
                 "play_date",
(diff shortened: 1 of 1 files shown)


gammaw:
@tuz666 yeah, I noticed it :) It took a while before I found that this PR is giving me the answer of why my local arcsi GUI has fallen apart. 

Show edits still don't seem to work though. It seems that `show_schema` is missing the fields `archive_lahmastore`, `archive_mixcloud` and `week` (marshmallow was complaining validation error until I added these fields to `show_schema`). I'm not sure where those changes sneaked in, maybe it was once adding the `only` keyword. Can you fix it on this branch before we merge it?

tuz666:
Nice catch, I didn't check all the use cases... Thanks for the feedback!

gammaw (approved):
@tuz666 thanks! This is another good example why we badly need some CI. arcsi is growing more and more complex, out reviews are not robust enough.

PR #139, opened 2021-12-17
Dev/couple of debug lines

gammaw:
It was a brute force/desperate debugging approach, which was helpful for me when I was trying to figure out what was going on in https://github.com/mmmnmnm/lahmacun_arcsi/pull/135, but since it's only active in debug mode (which is not default) I thought it may be sensible for the time being. Also it is not complete by any sense, I just kept adding lines until I found what I was looking for. We don't have to accept it as is, it may just trigger an initial discussion to find better solutions, but if you think it helps, let's add it for now. 

Heads up: the PR contains another PR (fix by tuz) and one change of parameter in `save_file`. If they turn out to be unnecessary, we can get rid of them when merging.


tuz666 (approved):
These new log lines are really helpful for debugging, I added some changes related to the CI's logging too, the aguni.log is saved as an artifact now, if pipeline fails.

PR #141, opened 2022-02-15
Quick fix for the legacy API calls

tuz666:
- In a previous merged commit I accidently removed the archived boolean from ShowDetailsSchema's items field. Sorry for that, I believed that we are getting the items of the given show from the /show/{slug}/archive endpoint
- Moved name duplications checker into utils

### DIFF
--- a/arcsi/api/item.py
+++ b/arcsi/api/item.py
@@ -18,6 +18,7 @@ from .utils import (
     dict_to_obj,
     media_path,
     save_file,
+    item_duplications_number
 )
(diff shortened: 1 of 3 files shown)


gammaw (approved):
Bravo! Code looks good and my tests also passed on dev. Thanks!

pvj:
Good job!

>Sorry for that, I believed that we are getting the items of the given show from the /show/{slug}/archive endpoint

I mean right now we don't but we aim to get the items separately in the future

PR #142, opened 2022-02-18
Dev/add error log for playlist assign

gammaw:
Fixed one failed logging behind a return and added more logging for playlist detection. Tested locally and on dev. Used to debug failing Azuracast upload for one show. Please review.

### DIFF
--- a/arcsi/handler/upload.py
+++ b/arcsi/handler/upload.py
@@ -251,8 +251,8 @@ class AzuraArchive(object):
             for playlist in r.json():
                 if playlist["name"] == self.playlist_name:
                     self.playlist_id = str(playlist["id"])
-                    return True
                     app.logger.debug("Add to playlist request returned {}".format(r.status_code))
+                    return True                
             app.logger.debug("ERROR: Couldn't find playlist ID in Azuracast response.")
             return False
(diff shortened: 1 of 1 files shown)


pvj (approved):

arcsi/handler/upload.py
```diff
@@ -251,8 +251,11 @@ def find_playlist_id(self,):
             for playlist in r.json():
                 if playlist["name"] == self.playlist_name:
                     self.playlist_id = str(playlist["id"])
-                    return True
+                    app.logger.debug("Add to playlist request returned {}".format(r.status_code))
```
pvj, line 254: One small idea here would be (really just cosmetic, shouldn't keep you from merging) to do `request returned {:d}` to suggest clearly that the `r.status_code` is an `Integer`

pvj (approved):

tuz666 (approved):
Looks good to me too!

arcsi/handler/upload.py
gammaw: Good idea. Could you add it when doing this https://github.com/mmmnmnm/lahmacun_arcsi/issues/114?

PR #152, opened 2022-07-26
Include Mixcloud URL field in page API

gammaw:
Aux PR for https://github.com/lahmacunradio/frontend/pull/272. Please review.


tuz666 (approved):

gammaw:
Thanks @tuz666 for the adding the frontend related API tests! Note: not only tests related to the original API change (`page`) but all missing ones (except one with `schedule_by?day` which didn't work for some reason and you left it commented out).

tuz666:
> Thanks @tuz666 for the adding the frontend related API tests! Note: not only tests related to the original API change (`page`) but all missing ones (except one with `schedule_by?day` which didn't work for some reason and you left it commented out). 

Yes, I don't know why that didn't work. I already tested it locally, got the same [] response what the CI had but locally it's worked. I just didn't want to delay the merge of this change that's why I commented it out.

Issue #126, opened 2021-11-05
Create new auth handling logic 


pvj:
As mentioned in #107 we are in the middle of creating auth handling views that lets us decouple from the current dependency that handles our auth routes & views. 

In order to remove that dep we have to adopt some auth logic that authenticates & authorises incoming requests.

For that we must change our api route handling: 
 - All requests are only valid if an active, existing token is passed -- preferably in the header
 - All invalid requests should be returned w/ 403 Forbidden HTTP status
 - Tokens should be generated by combination of current date + server-side secret + unique user data -- this should be handled by our newly selected auth provider plugin
 - Token generation and refresh, destruction etc. should be tied to user actions (login, logout etc.)
 - Tokens should be stored along w/ users

tuz666:
Resolved in [the following PR](https://github.com/lahmacunradio/arcsi/pull/156)

Issue #108, opened 2021-06-05
Auth token handling view


pvj:
We currently handle session tokens and user auth implicitly by just marking up the routes and checking these from the back-end but we should make sure that this is functionality for at least the admins on their user setting views so we can easily add / revoke them by just visiting the settings page 

Sister of #107

pvj:
Adding API key generation and handling to the front-end might be an extra we will only add if we sail past all the other must-haves. So for now I'll remove it from this milestone. Sorry~~~

Issue #107, opened 2021-06-05
Create new auth handling views


pvj:
The front-end client and the views handle and authenticate users and we have also roles and requirements in place. This allows to fine-tune who may access what. We should translate this functionality to not just views but also requests if possible by using the same session as the front-end client uses

 Sister of #108

pvj:
We planned how we will move forward w/ this. 

We will depend on `flask-praetorian` for handling role-based access & API auth. 

We will create token handling during login-flow parallel to the already existing one. 
  - We will recreate login view and templates, 
  - create login endpoints on the back-end, 
  - we *might* have to touch the models & our schemas while we are doing so as well,
  - we also will have to look into whether we may or may not be able to continue to use the already existing roles. If not we will have to add that logic again 
 
Once authentication & authorisation work with this flow we can compare it w/ the existing one, run them in parallel, verify that it doesn't break any existing route or functionality. 
Then we can add the other routes  as well (registration, logout, forgot pw, etc.), and finally deprecate the old auth-handling logic.

tuz666:
We've double checked it with @pvj, it's resolved now!

PR #153, opened 2022-07-27
Dev/#107 Authentication tokens for APIs

tuz666:
A long awaited missing piece of our Arcsi seems to be finished soon. (#107 , #108 )

Our first idea to implement authentication token handling was to start using flask-praetorian, which is a light-weight, API-oriented security package. (#126 ) As far as I remember we also thought that flask-security - what we are using currently - is a dead project and doesn't supported anymore.

I made some attempts to import flask-praetorian into our code parallelly with flask-security. During these process and research it turned out that flask-security also contains the features which we wanted to bring in with flask-praetorian. Also I found a package named flask-security-too, which is an independent well-maintained fork with high code coverage and nice documentation. After these news I stopped the experiments with flask-praetorian and started to work with flask-security.


entrypoint.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 python3 -m flask db upgrade
-gunicorn --worker-tmp-dir /dev/shm -w 1 --worker-class=gthread --threads 4 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level info "arcsi:create_app('../config.py')"
+gunicorn --reload --worker-tmp-dir /dev/shm -w 1 --worker-class=gthread --threads 4 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level info "arcsi:create_app('../config.py')"
```
pvj, line 3: This is for local work, or do you intend to patch the prod codebase as well? 

If needed for local work: We could look into a way to set this based on the runtime environment (local/dev/prod)

If this is intended, can you share the story?

docker-compose.yml
```diff
@@ -33,7 +33,7 @@ services:
         - app.env
         - db.env
     # there is no `depend_on: condition:` in compose service v3
-    entrypoint: sh -c 'sleep 60; /app/entrypoint.sh'
+    entrypoint: sh -c 'sleep 5; /app/entrypoint.sh'
```
pvj, line 36: Same as below. Is this an accident or intent?

config.template.py
```diff
@@ -19,6 +19,12 @@
 SECURITY_REGISTERABLE = True
 SECURITY_PASSWORD_SALT = "some_salt"
 
+# CONFIG USED FOR FLASK-PRAETORIAN
```
pvj, line 22: Nice setup, good names and defaults!

arcsi/view/archive.py
```diff
@@ -15,3 +16,33 @@ def archive():
...
+    #hashed_password = guard.hash_password(password)
+    #app.logger.error("password: " + password)
+    #app.logger.error("hashed_password: " + hashed_password)
+    #return jsonify(hashed_password)
+    user = guard.authenticate(username, password)
+    
+    ret = {"access_token": guard.encode_jwt_token(user)}
+    return (jsonify(ret), 200)
+
+@router.route("/api_key_nemadmin")
```
pvj, line 35: I like this way of testing behaviour by creating test endpoint but please push it with a simple standard name. A hunglish out of schema name is harder to use and easy to miss when cleaning up.

arcsi/api/show.py
```diff
@@ -88,15 +89,18 @@ def make_show(self, data, **kwargs):
...
     return shows_schema.dumps(get_shows())
 
 
 @arcsi.route("/show/all_without_items", methods=["GET"])
+@auth_required
 def list_shows_without_items():
     return shows_schedule_schema.dumps(get_shows())   
 
 @arcsi.route("/show/schedule", methods=["GET"])
+@auth_required
```
pvj, line 103: Here however I don't agree with this way of updating all the existing routes. 

This change is more related to the 2nd part of this feature, validation between the Arcsi api and the Arcsi website.

If you would like to test the api tokens and the permissions use 1 dummy copied from the existing apis or better yet just make some new api endpoints (which we might then keep as well) and test those. 

With the current approach, it's hard to figure out the next steps clearly and even harder to think what the ideal solution is going to be, that's why I'd ask for some revision on these.

arcsi/model/user.py
```diff
@@ -27,3 +28,20 @@ class User(db.Model, UserMixin):
 
     def __repr__(self):
         return "<Host {}>".format(self.name)
+
+class User2(db.Model, SQLAlchemyUserMixin):
```
pvj, line 32: This is a good solution to independently work with and develop the flask based authentication and authorisation! 

Are these columns defined by you or by some template? 

Because it seems like compared to the `User` model we are currently using the only addition by praetorian is `is_active` the only addition by us are the `butt_u` `butt_p` `shows` columns everything else mandatory is the same. So there should be no problem transforming current users then merging it back into User! 

Good job!

arcsi/model/user.py
```diff
@@ -27,3 +28,20 @@ class User(db.Model, UserMixin):
...
+    """
+
+    __tablename__ = "users2"
+    id = db.Column(db.Integer, primary_key=True)
+    username = db.Column(db.Text, unique=True)
+    hashed_password = db.Column(db.Text)
+    roles = db.Column(db.Text)
+    is_active = db.Column(db.Boolean, default=True, server_default="true")
+
+    def is_valid(self):
```
pvj, line 46: Small nitpick, sorry but I am curious:

Did you choose `is_valid` because `is_active` is an existing class method you don't want to mess with or just by chance?

If just by chance could I ask you to use `is_active`? I have experienced many times that different method names to argument names (especially if it's a one-line getter) can create confusion. Which is the field name, which is the method, `valid` or `active`... are they equivalent? What if we want to have a separate `valid` property later on, etc.

pvj (commented):
Commented your draft Solid first steps and good test db and test key set-up. I hope to help when you continue

arcsi/model/user.py
```diff
@@ -27,3 +28,20 @@ class User(db.Model, UserMixin):
...
+    well defined
+    """
+
+    __tablename__ = "users2"
+    id = db.Column(db.Integer, primary_key=True)
+    username = db.Column(db.Text, unique=True)
+    hashed_password = db.Column(db.Text)
+    roles = db.Column(db.Text)
+    is_active = db.Column(db.Boolean, default=True, server_default="true")
+
```
pvj, line 45: Where is the api key column handled?

entrypoint.sh
tuz666: This is just for testing purposes. Since this is just a draft PR, it was more comfortable to push it, than stashing it in and out every time.

docker-compose.yml
tuz666: I think neither of them, reasoning is in my previous comment.

config.template.py
tuz666: As I remember we did this together, thanks for the help :)

arcsi/view/archive.py
tuz666: These two APIs are really just for testing the functionalities, also they are containing hardcoded values.

arcsi/api/show.py
tuz666: I added these new decorators on the show related APIs in order to demonstrate praetorian's features. I added the auth_required() for the GET requests and roles_required() for POST and DELETE requests, seemed to me as a probably right implementation, but it's still a draft PR, so I'm happy to hear your ideas :)

arcsi/model/user.py
tuz666: Yes, I had some problems with the flask-security and flask-praetorian needed fields and methods. First I tried to copy the original User model (with name and password for example) and give back the flask-praetorian needed fields as a function (username() and hashed_password() functions gave back the value of name and password fields) but it turned out that these fields need to be exactly defined as username and hashed_password.

arcsi/model/user.py
tuz666: As I remember I just copied this lines from the flask-praetorian's example from [here](https://github.com/dusktreader/flask-praetorian/blob/master/example/basic_with_sqlalchemy_mixin.py)

arcsi/model/user.py
tuz666: In the same example I mentioned before I didn't see any field for it. Also I'm not sure about storing it in the db is a good idea, because it's going to be expired after 48 hours (or the time we set up in the config). I thought it is easier that if we need it we will just call the API (the proper, not the current hardcoded one) and it gives back the up-to-date token for us.

arcsi/templates/user/edit.html
```diff
@@ -5,7 +5,7 @@ <h2>{% block title %}Edit my user{% endblock %}</h2>
 {% endblock %}
 
 {% block content %}
-<form action="{{ url_for('arcsi.user', id=user.id) }}" method="post" enctype="multipart/form-data">
+<form action="{{ url_for('arcsi.view_user', id=user.id) }}" method="post" enctype="multipart/form-data">
```
gammaw, line 8: @tuz666 just curious, why this change? Was it typo here?

arcsi/view/item.py
```diff
@@ -1,16 +1,17 @@
...
 from flask_security import roles_accepted
 
 from arcsi.view import router
 
 
 @router.route("/item/all")
+@roles_accepted("admin", "host", "guest")
 def list_items():
-    result = requests.get(app.config["APP_BASE_URL"] + url_for("arcsi.list_items"))
+    result = requests.get(app.config["APP_BASE_URL"] + url_for("arcsi.list_items"), headers = {"Authentication-Token": current_user.get_auth_token()})
```
gammaw, line 14: @tuz666 why does this work and other APIs don't when used internally? I thought that's the point of separation between login and auth. Or is there a difference if you call a GET/POST API or if you call it from a view? Sorry if the question is something we discussed already, I can't recall it now.

test/empty_dump.sql
```diff
@@ -284,7 +284,7 @@ COPY public.roles_users (user_id, role_id) FROM stdin;
 --
 
 COPY public.users (id, name, email, butt_user, butt_pw, password, active) FROM stdin;
-1	testuser	testuser@lahmacun.hu	buttuser	buttpw	testpassword	t
+1	testuser	testuser@lahmacun.hu	buttuser	buttpw	$2b$12$bz/hCRWRMjNeHGr66wYZO.a/HHhenAHZaLECKmVYPFLSVTzvC/l9a	t
```
gammaw, line 287: @tuz666 sure it's not a plain text public password which grants the srongest access rights? I'm note I expected it to go somehow via GH actions secrets, but you're the expert.

gammaw (commented):
@tuz666 thanks for the bold changes! Looks just like we discussed, I just had a couple of questions for clarification before I approve. Can you look into them?

arcsi/api/item.py
```diff
@@ -126,8 +114,89 @@ def view_item(id):
...
+
+
+@arcsi.route("/item/<id>/edit", methods=["POST"])
+@login_required
+@roles_required("admin")
+def edit_item(id):
+    return _edit_item(id)
+
+
+@arcsi.route("/item/<id>/edit_api", methods=["POST"])
```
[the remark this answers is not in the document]
tuz666: The only reason I forgot about it, but gonna upload them in a new commit

arcsi/api/show.py
```diff
@@ -301,118 +448,3 @@ def edit_show(id):
         return make_response(
             jsonify(show_partial_schema.dump(show)), 200, headers
         )
-
```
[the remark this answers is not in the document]
tuz666: Yes I also saw this problem, I thought if I did it in a separate commit then git could handle it, unfortunately not :(
I'm gonna revert this change then and do it in a separate PR.

arcsi/api/user.py
```diff
@@ -53,3 +48,22 @@ def view_user(id):
         return jsonify(user_details_schema.dump(user))
     else:
         return make_response(jsonify("Could not find user", 404, headers))
+
+
+@arcsi.route("/users/get_api_token", methods=["POST"])
```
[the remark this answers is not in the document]
tuz666: Forgot about this too :(

arcsi/api/utils.py
```diff
@@ -6,7 +6,7 @@
 from arcsi.model.item import Item
 from flask import current_app as app
 from slugify import slugify
-from werkzeug import secure_filename
+from werkzeug.utils import secure_filename
```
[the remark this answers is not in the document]
tuz666: You got me with this one, as I remember I tried out a newer version of flask-security which needed newer werkzeug as its dependency, and it caused ImportError in that case.
Since we won't do any version update here, maybe it is better just to revert this change, and do it when it is necessary.

arcsi/templates/user/edit.html
tuz666: Exactly, you can easily test it just try to edit your profile in the prod and dev page.

arcsi/view/item.py
tuz666: Yes, there is a difference.
If you are logged in with a user it is possible to generate tokens (and add them to the headers) for GET requests which are used for view functions.
But you can't add headers to forms and that's why I had to create separate POST APIs for internal and external use.

test/empty_dump.sql
tuz666: This empty db is used only for the CI and that is 'testpassword' after hashing, it was needed in order to get the API token.
If you give me access to set up GH Actions secrets then I will try to store these stuff there :)

gammaw: Oh, I see, you're right, it's not sensitive then (cause this user/pwsd is not in our public arcsi instances).

gammaw (approved):
Thanks for the comments and the changes, especially, the Swagger updates, which was additional work! Let's start step 1 then (arcsi release w/ only POST APIs being protected)!

tuz666:
Thanks for the review @gammaw , I will push the green button then :)

PR #156, opened 2022-10-11
Dev/#107 Authentication tokens for APIs pt. 2

tuz666:
We separated the Authentication tokens for APIs task into two PRs in order to provide time for our frontend and mobile app for a smoother uplift:
1. During the first PR (#153) we introduced the auth_token handling and protected our POST and DELETE APIs.
2. Now we gonna make our GET APIs protected too.

etc:
- I did a little refactoring with the duplicated add/edit_item/show functions, now it doesn't mess up the whole changes view.
- Sometimes the CI test failed on the item upload, because it reached the maximum 10 sec response time, I increased this maximum value to 12.5 sec
- Updated the annotations on Arcsi's view functions (if someone register he/she/they can have a look on it), and fixed the episode listen/download api?


arcsi/api/item.py
```diff
@@ -457,6 +298,7 @@ def listen_play_file(id):
 
 
 @arcsi.route("/item/<id>/download", methods=["GET"])
+@roles_required("admin", "host", "guest")
```
gammaw, line 301: @tuz666 why is that? We could use `@auth_token_required` here too, no?

arcsi/api/item.py
```diff
@@ -277,173 +280,11 @@ def add_item():
...
-                jsonify(item_schema.dump(new_item)),
-                200,
-                headers,
-            )
-        else:
-            return "Some error happened, check server logs for details. Note that your media may have been uploaded (to DO and/or Azurcast)."
-
+    return add_item()
 
 @arcsi.route("item/<id>/listen", methods=["GET"])
```
gammaw, line 289: @tuz666 isn't `@auth_token_required` missing here?

arcsi/api/item.py
gammaw: This one is clear. `@roles_required` only means that we don't assume if it's logged in user or an API call. Note: on site registrations will not assign a role (not even `guest`) so this one is a real restriction (do `admin`, `host` or `guest`).

tuz666: This was also a good question, I thought this means that if you have any of the listed roles than have the authentication, but it turned out that you should have each of them.

arcsi/api/item.py
tuz666: Yes, I forgot about it, nice catch!

gammaw (approved):
Nice clean-up! And clarification of the exact meaning of the different auth annotations. Also, thanks for the small player adjustment.

PR #157, opened 2022-10-30
Changing flask-security to flask-security-too

tuz666:
The currently used [flask-security](https://flask-security.readthedocs.io/en/3.0.0/) package is no longer supported (latest change is 2,5 years old), [flask-security-too](https://flask-security-too.readthedocs.io/en/stable/) is the maintained up-to-date version of it.

Previously we talked about @gammaw that introducing the authentication made the frontend page is a bit slower.
This new security package offers some performance boost with version [3.1.0](https://github.com/Flask-Middleware/flask-security/blob/master/CHANGES.rst#version-310) and [3.2.0](https://github.com/Flask-Middleware/flask-security/blob/master/CHANGES.rst#version-320), for these uplifts we need a [modification](https://github.com/lahmacunradio/arcsi/pull/157/files/db80fc450235b0191d312f721a6aa68717fdfd73#diff-b221de923024f3e630e3d372fae43d57a72164cbe70f71b0dd9df23b36606983) which looks like a DB change [but it's not that](https://github.com/lahmacunradio/arcsi/pull/157/files/db80fc450235b0191d312f721a6aa68717fdfd73#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552).

We agreed that we should do some performance measurement before these upgrades:

- @gammaw suggested to do a profiling on the front end page to see exactly how much time each transaction takes
- I recommended to extend our CI with a robustness/benchmark test which can simulate a big load of incoming GET requests and measure the time it takes.

Because of the previously mentioned reasons I moved back to version 3.0.2, since this PR will be just about introducing the new flask-security package and probably the CI upgrade.

### DIFF
--- a/arcsi/model/user.py
+++ b/arcsi/model/user.py
@@ -14,7 +14,7 @@ class User(db.Model, UserMixin):
     butt_user = db.Column(db.String())
     butt_pw = db.Column(db.String(128))
     roles = db.relationship(
-        "Role", secondary=roles_users, backref=db.backref("users")
+        "Role", secondary=roles_users, backref=db.backref("users"), lazy="dynamic"
     )
     shows = db.relationship(
         "Show",
--- a/arcsi/model/user.py
+++ b/arcsi/model/user.py
@@ -14,7 +14,7 @@ class User(db.Model, UserMixin):
     butt_user = db.Column(db.String())
     butt_pw = db.Column(db.String(128))
     roles = db.relationship(
-        "Role", secondary=roles_users, backref=db.backref("users"), lazy="dynamic"
+        "Role", secondary=roles_users, backref=db.backref("users")
     )
     shows = db.relationship(
         "Show",
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ Flask-Mail==0.9.1
 flask-marshmallow==0.12.0
 Flask-Migrate==2.5.2
 Flask-Principal==0.4.0
-Flask-Security-Too==3.1.0
+Flask-Security-Too==3.1.0rc1
 Flask-SQLAlchemy==2.4.2
 Flask-WTF==0.14.3
 flowjs==0.1.1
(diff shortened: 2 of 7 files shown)


requirements.txt
```diff
@@ -24,7 +24,7 @@ Flask-Mail==0.9.1
 flask-marshmallow==0.12.0
 Flask-Migrate==2.5.2
 Flask-Principal==0.4.0
-Flask-Security==3.0.0
+Flask-Security-Too==3.2.0
```
gammaw, line 27: @tuz666 do we expect a performance boost from the bump? Did you deploy this branch on dev?

arcsi/model/user.py
```diff
@@ -14,7 +14,7 @@ class User(db.Model, UserMixin):
     butt_user = db.Column(db.String())
     butt_pw = db.Column(db.String(128))
     roles = db.relationship(
-        "Role", secondary=roles_users, backref=db.backref("users"), lazy="dynamic"
```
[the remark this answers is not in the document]
tuz666: Sorry, I should mark this PR as draft, or at least write a proper description about the modifications.
This change was suggested in new flask-security's [change log](https://github.com/Flask-Middleware/flask-security/blob/master/CHANGES.rst#version-310).
It was also a good test for the CI as you can see if I tried to update only for version 3.0.2 it went fine.
In case when I wanted to update for 3.2.0 it failed because it missed this change, with this change it passed again

requirements.txt
tuz666: Yes, you are right, both version uplift 3.1.0 and 3.2.0 promise a performance boost, and only that little relationship change needed. I tried to create the DB migration script but it turned out this modification won't cause any direct DB change, also we had a conversation about the topic with @pvj and he confirmed this too.

gammaw: Thanks!

> Did you deploy this branch on dev?

And how about this? I'd do this first before merging.

tuz666: I deployed it on the dev now, not sure about the situation of the frontend staging, I reach out to you directly.

PR #159, opened 2022-11-20
New endpoint for analytics

tuz666:
Added 2 new API endpoints which returns the sum of the uploaded items during the last x days/weeks from the given date.
Introduced new statistics related view pages on Arcsi UI, to be able to reach these info without Postman requests.
Updated the Swagger docs according to this.
Also added some constraints on API URL's variables, in order to prevent incorrect API calls (changes in item.py, show.py, user.py).

### DIFF
--- a/arcsi/api/data.py
+++ b/arcsi/api/data.py
@@ -22,7 +22,7 @@ def uploaded_episodes_in_last_x_days():
     episodes_count = Item.query.filter(
         given_date - timedelta(days=last_days) <= Item.play_date
         ).filter(Item.play_date <= datetime.today()
-        ).filter(Item.archived == True).count()
+        ).count()
     ret = {"uploaded_episodes_in_last_%d_days" % (last_days): episodes_count}
     return make_response(jsonify(ret), 200, headers)
 
@@ -40,6 +40,6 @@ def uploaded_episodes_in_last_x_weeks():
     episodes_count = Item.query.filter(
         given_date - timedelta(weeks=last_weeks) <= Item.play_date
         ).filter(Item.play_date <= datetime.today()
-        ).filter(Item.archived == True).count()
+        ).count()
     ret = {"uploaded_episodes_in_last_%d_weeks" % (last_weeks): episodes_count}
     return make_response(jsonify(ret), 200, headers)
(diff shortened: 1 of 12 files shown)


arcsi/api/data.py
```diff
@@ -0,0 +1,45 @@
...
+    given_date = datetime.strptime(from_date, '%Y-%m-%d')
+    if (given_date > datetime.today()):
+        return make_response("Back to the future?", 418, headers)
+    last_days = int(date_metadata['last_days'])
+    if (last_days < 0):
+        return make_response("Please, be postive!", 400, headers)
+    episodes_count = Item.query.filter(
+        given_date - timedelta(days=last_days) <= Item.play_date
+        ).filter(Item.play_date <= datetime.today()
+        ).filter(Item.archived == True).count()
```
gammaw, line 25: @tuz666 it seems we don't count live episods that we haven't archived yet. I think we should, cause they have been aired too.

gammaw (changes requested):

arcsi/api/item.py
```diff
@@ -97,7 +97,7 @@ def list_items_latest():
     return items_archive_schema.dumps(items.items)
 
 
-@arcsi.route("/item/<id>", methods=["GET"])
+@arcsi.route("/item/<int:id>", methods=["GET"])
```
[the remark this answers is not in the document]
tuz666: Yepp, I will do that way next time!

arcsi/api/data.py
```diff
@@ -22,7 +22,7 @@ def uploaded_episodes_in_last_x_days():
     episodes_count = Item.query.filter(
         given_date - timedelta(days=last_days) <= Item.play_date
         ).filter(Item.play_date <= datetime.today()
-        ).filter(Item.archived == True).count()
```
gammaw: @tuz666 elegant! Adding features by deleting code ;)

gammaw (approved):
Thanks @tuz666 for the fast & quality implementation of this!

arcsi/api/data.py
tuz666: :DD thanks for review!!

PR #160, opened 2022-11-27
Add missing imports

tuz666:
Background:
During a previously [merged PR](https://github.com/lahmacunradio/arcsi/pull/153) I removed some unused imports, unfortunately it caused a failure during item/show edit/upload debugging, which was merged with [this PR](https://github.com/lahmacunradio/arcsi/pull/139).

### DIFF
--- a/arcsi/api/item.py
+++ b/arcsi/api/item.py
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta
 
 from flask import jsonify, make_response, request, redirect
+from flask import current_app as app
 from flask_security import auth_token_required, roles_required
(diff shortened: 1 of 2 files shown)


gammaw (approved):
Thanks for the quick fix!

I wonder how this error was not caught by our CI?

tuz666:
Thanks for the quick review @gammaw !
I also wondered, why it didn't cause a failure, since the editing of shows/episodes are tested there too, probably it's because I only did positive testcases. 
Anyway I created an [issue](https://github.com/lahmacunradio/arcsi/issues/161) about the missing negative testcases

PR #166, opened 2023-01-22
Taggable shows & episodes

pvj:
Created a `Tag` model, table in our db, serialisation schema for the API, and a few html forms to create and update and tags on shows (or even episodes). 

I also tried to keep a bit more self-restraint and keep most of the functionality simple and sorted. Any extra functions are written to the utilities


tuz666 (approved):
Thanks for this change tooo @pvj !!
I updated the CI tests and tested it locally, everything looked fine!

tuz666:
I think this issue was solved in this PR
https://github.com/lahmacunradio/arcsi/issues/124

PR #167, opened 2023-01-29
Social media url for shows

pvj:
Hi

I have tried to refactor our show models, show Api serial schemas and the test-cases / Api docs.

We never used the `archive_mixcloud_base_url` field for the shows and it seemed easy to rename that so we can introduce a similar but orthogonal functionality: displaying external links / social media Urls for the shows.

Note: I tried to not touch the item fields w/ mixcloud but I might have removed some!


tuz666 (approved):
Everything looks fine, thanks for the PR!
I extended the changes for the items table as well and adjusted the CI testcases too.
Quick notes:
I modified to use the little bit faster `list_shows_without_items` instead of the plain `list_shows` API for the edit item page and added id based ordering on the `latest` API.

PR #168, opened 2023-01-29
Contact field for shows

pvj:
This branch contains work on the field contact_address 

I thought about the models and where to add this column and went with shows instead of hosts. It's good if each show has a fixed field and then the actual addresses or how that is accessed can be improved / changed later. And logically, we wanted to have a feature where shows can be messaged, not hosts, so it seemed like the better option. 

It is a simple char field but the schema has the field `Email` set. 

Crud templates have been updated. 

Db schema migrations was created / verified


pvj:
Note: As these pull reqs all created migrations they will have to be amended so the up-down-grade paths are linear

tuz666 (approved):
Just testing the new branch merging limitations
P.S. I checked it locally and it works fineee!

Issue #144, opened 2022-03-08
Restrict item name hashing to show scope


gammaw:
Description: item name will be hashed in case of match w/ existing item in other show (e.g., 18. rész in Mélyvágás vs Üdcsi).

Thanks @tuz666 for the finding! I'll check.

tuz666:
done

Issue #113, opened 2021-07-15
Optimise response with nested materials


pvj:
Some endpoints have quite heavy payloads -- a show endpoint shouldn't load all the related episodes and all the metadata for them, and download all data. 

Also if I have a list of episodes w/ images to download on a list page we should request the original files to download each time.

tuz666:
Closed it with @pvj !

PR #172, opened 2023-06-02
Minor Arcsi improvements

tuz666:
- Fixed initial value of `active` field on Show edit page
https://github.com/lahmacunradio/arcsi/issues/174
- Fixed the initial values for `tags` field on Show and Episode edit page
https://github.com/lahmacunradio/arcsi/issues/174
- Removed `contact_address` from frontend used APIs
- Fixed the `item_duplications_number` utility function to only check the duplications between the episodes of the given show, instead of the whole episode db
https://github.com/lahmacunradio/arcsi/issues/144
- Improved `/show/<show_slug>/page` API to be more versatile
https://github.com/lahmacunradio/arcsi/issues/113
https://github.com/lahmacunradio/frontend/issues/288
- Inceased maximum size of `social_base_url` in DB and renamed it to 'external_url'
https://github.com/lahmacunradio/arcsi/issues/173
- Updated CI tests to use authentication tokens with both guest and admin roles


arcsi/api/show.py
```diff
@@ -389,30 +389,52 @@ def view_show_archive(show_slug):
 @auth_token_required
 def view_show_page(show_slug):
     do = DoArchive()
+    archived = request.args.get('archived', True, type=lambda v: v.lower() == 'true')
```
pvj, line 392: Thanks for implementing the idea of these parameter controlled dynamic views. For a couple of reasons, which I will try to convince you about please consider to change almost the whole logic of this part of your changes, or pull these out of this PR, and have a separate branch where we can work on only this as a focus.

First, these two variables are used for a clear goal here, to decide what data to send back with the view. That's a good intent! The way it is defined is bothering me. I will try to explain what parts need to be thought about and will try to include how I would go about and what we benefit.

0, `Archived` and `latest` might be useful in the Show api in other parts of the Api. I would place these variables out from method scope. Then all public endpoints are aware of the same arguments and state.
1, There should be one method in the module that does what these lines. I'm looking at the `utils` pack and imagine I would add there a short function that;
 - imports the request context, 
 - takes a string we want as input and 
 - lists all argument keys searching for it. 
(It can return a boolean or the key or the value, it isn't that important.) I would look into `after_request` so we don't have to call parameter parsing on each endpoint and that has benefits like simplicity in the HTTP handling part, or that we can apply these for all views in general
2, Here we are returning the Boolean. It is a weird way to do I would definitely think of just having `?archived` and handling that (this `lambda` implies to me that we handle `?archived=True` is that correct or not?) 

Now don't take me wrong, I believe your are exercising the good practice of gardening where you always improve things in small bits as you find them that I appreciate a lot. I would like to only highlight that internal improvements of the code have to be a top consideration also.

arcsi/api/show.py
```diff
@@ -389,30 +389,52 @@ def view_show_archive(show_slug):
...
-            < datetime.today()
-        ]
-        serial_show["items"]=show_items
-        for item in serial_show["items"]:
-            item["image_url"] = do.download(
-                show.archive_lahmastore_base_url, item["image_url"]
-            )
-            item["name_slug"]=normalise(item["name"])
-        return serial_show
+        if (archived):
```
pvj, line 402: Syntax small note: please only add paren if there are multiple logical operations

arcsi/api/show.py
```diff
@@ -389,30 +389,52 @@ def view_show_archive(show_slug):
...
+                serial_show["items"]["name_slug"] = normalise(serial_show["items"]["name"])
+                return serial_show
+            else:
+                for item in serial_show["items"]:
+                    item["image_url"] = do.download(
+                        show.archive_lahmastore_base_url, item["image_url"]
+                    )
+                    item["name_slug"] = normalise(item["name"])
+                return serial_show
+
```
pvj, line 437: This whole part is the type of pattern that exists currently elsewhere in the Api (I wrote a lot of it as well) but we should not replicate it instead trying to move away. 

The problem with this type of view construction is that is lazy in the bad way. 
1, It makes us do the work twice but also for the server we make it execute twice. 
2, Also in a bad way lazy for data structures. We reconstruct each db query, media downloads, python functions etc. multiple times. 
3, While it gives a shortcut when writing or thinking about it, it is hard to extend to other parameters, or change the relations, or the data we want to get. 

The nested `if-else` statements with `archived` and `latest` are a red alert for me in such situations, they alert that it's a pattern that will be dependent and coupled and that makes it easy to break, hard to maintain.

arcsi/api/show.py
```diff
@@ -389,30 +389,52 @@ def view_show_archive(show_slug):
...
-        return serial_show
+        if (archived):
+            show_items = [
+                show_item
+                for show_item in serial_show["items"]
+                if (show_item.get("archived") &
+                    (datetime.strptime(show_item.get("play_date"), "%Y-%m-%d") + timedelta(days=1) < datetime.today()))
+            ]
+            if (latest):
+                show_items = sorted(show_items, key=lambda x: x["play_date"], reverse=True)
```
pvj, line 410: Sorted show_items should be defined once. Function could capture this procedure so it is computed once.

Currently this is defined twice but the second one is just a copy-paste below a different statement.

arcsi/api/show.py
```diff
@@ -389,30 +389,52 @@ def view_show_archive(show_slug):
...
+                serial_show["items"] = show_items[0]
+                serial_show["items"]["image_url"] = do.download(show.archive_lahmastore_base_url, serial_show["items"]["image_url"])
+                serial_show["items"]["name_slug"] = normalise(serial_show["items"]["name"])
+                return serial_show
+            else:
+                serial_show["items"] = show_items
```
pvj, line 416: The whole logic of this part is reversed -- it calculates the differences and after that goes through the common parts. 

First we should set up the common parts of `serial_show` then in one step make any adjustments. 

The benefit is that the most common view is handled first and any adjustments can be independently reworked, easily extended or disabled without fears of breaking the default view.

arcsi/api/show.py
```diff
@@ -389,30 +389,52 @@ def view_show_archive(show_slug):
...
+                    item["name_slug"] = normalise(item["name"])
+                return serial_show
+        else:
+            if (latest):
+                serial_show["items"] = sorted(serial_show["items"], key=lambda x: x["play_date"], reverse=True)
+                serial_show["items"] = serial_show["items"][0]
+                serial_show["items"]["image_url"] = do.download(show.archive_lahmastore_base_url, serial_show["items"]["image_url"])
+                serial_show["items"]["name_slug"] = normalise(serial_show["items"]["name"])
+                return serial_show
+            else:
```
pvj, line 430: This is I guess the default path right? This should be captured to always reliably run and return at first. 

When we do any modification to this query (here: `latest` and `archived`) the server can just fall back and return the default list above in case there was an issue with modification

arcsi/api/utils.py
```diff
@@ -159,6 +159,8 @@ def get_shows():
     return shows
 
 def item_duplications_number(item):
-    name_occurrence = int(db.session.query(db.func.count()).filter(Item.name == item.name, Item.number == item.number).scalar())
+    show_items = Show.query.filter(Show.id == item.shows[0].id).first().items
+    show_items_with_same_name = show_items.filter_by(name=item.name)
```
pvj, line 163: Thanks for breaking this long line up into smaller parts!

One naming nitpick. We are in the scope of this function there is no need to add the `show_` prefix. Would you think otherwise? 

Also it is a function that can be read on one screen so it is readable that it is a Show model's query filter.

arcsi/api/utils.py
```diff
@@ -159,6 +159,8 @@ def get_shows():
...
+    name_occurrence = show_items_with_same_name.count()
     app.logger.error("Name_occurence (duplicate detection): {}".format(name_occurrence))
-    return name_occurrence
+    return 0 < name_occurrence
```
pvj, line 166: I mean you probably wanted to have a more exact boolean return here and get `True` or `False` instead of `1` and `0` 

But if you think about this adjustment in broad terms it does matter a bit since now you changed the return type from integer to boolean so if it is used somewhere as a number not just a logical statement this change could break the application.

It can be a good change to make it return boolean but on changes like this we either should add inline documentation or more details on the PR description about this.

pvj (changes requested):
Thanks! I have two main issues please take your time to review those, and some which are just syntax or preference. 

Thanks for the pushes and the PR even if my comments are long the main point is whenever you push anything you're ready with is the greatest choice regardless of the discussion happening in here. 

Keep up the noise work thanks!

arcsi/api/item.py
```diff
@@ -169,6 +169,10 @@ def add_item():
         tags = (
             get_or_create( Tag, display_name=tag.display_name, clean_name=normalise(tag.display_name) ) for tag in item_metadata.tags
         )
+
+        #Check for duplicate files
```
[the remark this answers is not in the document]
tuz666: Moved it back to its original place, the issue what I've seen (it always returned at least the new_item) only occured during the local testing

arcsi/api/show.py
tuz666: I did the refactor for this ugly if-else tree, hope you will like the current version too

arcsi/api/show.py
tuz666: After we discussed this part I can agree with you, I added a little workaround to replace the previously used default values (`archived=True` and `latest=False`, they are needed because the frontend already using this API without any params)

arcsi/api/show.py
```diff
@@ -389,29 +389,28 @@ def view_show_archive(show_slug):
 @auth_token_required
 def view_show_page(show_slug):
     do = DoArchive()
+    filterArray = {}
```
pvj, line 392: small syntactic nitpick: 
 - you've used camelcase for these vars but please always use the pythonic schema: `filter_array`
 - An `array` is a list `[]` this is a `dict`. Could be `filter_dict`?

Here you may not even need that second part and just call it `filter`.

arcsi/api/show.py
```diff
@@ -389,29 +389,28 @@ def view_show_archive(show_slug):
...
+    filterList = request.args.getlist('filter')
```
pvj, line 393: Good use of `getlist` here!

arcsi/api/show.py
```diff
@@ -389,29 +389,28 @@ def view_show_archive(show_slug):
...
+    if len(filterList) == 1 and ',' in filterList[0]:
+        filterArray['filters'] = comma_separated_params_to_list(filterList[0])
+    else:
+        filterArray['filters'] = filterList 
+    # TODO workaround for current frontend usage
+    archived = 'archived' in filterArray['filters']
+    if archived == None:
+        if "https://lahmacun.hu" in request.environ.get('HTTP_ORIGIN'):
```
pvj, line 401: 🛖

arcsi/api/utils.py
```diff
@@ -158,7 +159,45 @@ def get_shows():
...
+        
+def get_show_items(show, items, latest):
+    do = DoArchive()
+    if latest:
+        items = items[0]
+        items["image_url"] = do.download(show.archive_lahmastore_base_url, items["image_url"])
+        items["name_slug"] = normalise(items["name"])
+        return items
+    else:
+        items = items
```
pvj, line 197: This may be not necessary you can just loop through the incoming argument of `items` and return the modified version of it.

pvj (dismissed):
Really good job! The requested changes have been resolved in a really nice fashion. You have improved not just the functionality and features of arcsi but our code-base as well.  

There is one small syntax-change I ask you to make, otherwise it looks good to go. 

Just let's remember that this PR includes a db migration as well!

arcsi/api/utils.py
tuz666: true 😅

arcsi/api/show.py
tuz666: yeppp, thanks for the review, I could simply use the filter_list here, I don't have to create another variable

arcsi/api/show.py
tuz666: thanks 💯

arcsi/api/show.py
tuz666: until this workaround is here, frontend can't use this query to get the future episodes (filters doesn't contain `archived`)

gammaw (approved):
Disclaimer: only reviewed the last two commits (since the last approve), which are 1) external URL var rename, 2) cleanup & ver bump. 

Thanks!

PR #175, opened 2023-06-20
Remove temporary API workaround

tuz666:
In order to be sync with the frontend we had to introduce a workaround in the previously merged PR (#172)

The changed API was already used by the frontend without any `filters` and the expected response was a filtered (without upcoming episodes) `items` field under the returned `show`.

Since the `archived` flag is added to the frontend code (in this PR https://github.com/lahmacunradio/frontend/pull/291) where it is needed, we can remove this workaround which added the `archived` flag in case the request was coming from the frontend

Connecting issues:
- https://github.com/lahmacunradio/frontend/issues/288
- https://github.com/lahmacunradio/frontend/issues/289


gammaw (approved):
Go go!

PR #177, opened 2023-08-23
Using tmpreaper utility to maintain the tmp files

pvj:
**Why?**
We tend to write to tmp storage any file that is posted to arcsi which may fill the disk before we ever restart the system. Manually checking and deleting files is cumbersome and may result in errors (an open file is forcibly deleted)

**What has changed?**
A proven, old, stable utility for this situation has been added to our image -- please read [the manual for tmpreader](https://manpages.ubuntu.com/manpages/lunar/en/man8/tmpreaper.8.html) for details on how and what it can do

Also added a config file that specifies that files not touched within 21 days should be removed. The default directory where tmpreaper works is `/tmp/` so there is no need to set anything else. 

**How does it work?**
Upon install tmpreaper adds a job to `/etc/cron.daily` which uses the config file at `/etc/tmpreaper.conf` 

**Tests**
Ran this on sauce servers since I could not access prod instances. If merged please check it on prod as well however due to the minimal settings and single usecase I expect it will work all right.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,11 @@ WORKDIR /app
 RUN apt-get update \
     && apt-get install gcc g++ \
     libblas-dev libffi-dev liblapack-dev libopenblas-dev libpq-dev \
-    musl-dev postgresql -y \
+    musl-dev postgresql tmpreaper -y \
     && apt-get clean \
     && pip3 install -r requirements.txt
+
+ADD infra/tmpreaper.conf /etc/tmpreaper.conf
 
 ADD . /app
 
--- a/infra/tmpreaper.conf
+++ b/infra/tmpreaper.conf
@@ -41,7 +41,7 @@ SHOWWARNING=false
 #       extra options that are passed to tmpreaper, e.g. --all
 
 # uncomment and change the next line to overrule the /etc/default/rcS value
-TMPREAPER_TIME=21d
+TMPREAPER_TIME=7d
 
 TMPREAPER_PROTECT_EXTRA=''
 TMPREAPER_DIRS='/tmp/.'
--- a/infra/tmpreaper.conf
+++ b/infra/tmpreaper.conf
@@ -1,0 +1,49 @@
+# tmpreaper.conf
+# - local configuration for tmpreaper's daily run
+#
+# This is only used if /etc/cron.daily/tmpreaper was also updated,
+# i.e. there's a line ". /etc/tmpreaper.conf" in that file.
+# The shell code that used to be here (pre version 1.6.7) is now
+# in the cron.daily script.
+
+# Remove the next line if you understand the possible security implications of
+# having tmpreaper run automatically;
+# see /usr/share/doc/tmpreaper/README.security.gz
+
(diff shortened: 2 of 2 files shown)


tuz666 (approved):
Thanks for the change and the nice description!
It looks clear and good to me!

gammaw:
Thanks @pvj! It's a much needed feature! I look forward to checking it. I hope to be able to do it in the next 1-2 weeks.

gammaw:
@pvj @tuz666 sorry guys, I overlooked Endre's approval. I was gonna validate it first before I approve, but it's good that you reviewed and merged it already. I will validate the feature in operation as it's already deployed. Thanks Jani again!

gammaw:
@pvj @tuz666 I had a look at this again, and I'm not sure about a couple of things: 

- If it's a Docker-based solution, why there is no change to the docker-compose file? Is it supposed to run as a standalone service? 
- Specifically, if tmpreaper is a new Docker service (the new Dockerfile suggests that), how would it have access to the arcsi's tmp directory? There are several options for that, but I don't see how this PR addresses the issue.

tuz666:
Hi @gammaw ,
Let me answer your questions here:

> If it's a Docker-based solution, why there is no change to the docker-compose file? Is it supposed to run as a standalone service?

It's an Ubuntu-based solution and we do have an indirect impact on the [docker-compose](https://github.com/lahmacunradio/arcsi/blob/master/docker-compose.yml#L23) file, because we added a line into the web container's [Dockerfile](https://github.com/lahmacunradio/arcsi/pull/177/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R12) to install the tmpreaper package. After the installation we will have the cleanup script at the following location: `/etc/cron.daily/tmpreape`r and there will be a configuration file here: `/etc/tmpreaper.conf` what we can [overwrite](https://github.com/lahmacunradio/arcsi/pull/177/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R16) with the added [file](https://github.com/lahmacunradio/arcsi/pull/177/files#diff-f2d4d677abd8de15baff35a12f09ebec8ed370bee8c2fcf9b2f53ccedaff0519)

> Specifically, if tmpreaper is a new Docker service (the new Dockerfile suggests that), how would it have access to the arcsi's tmp directory? There are several options for that, but I don't see how this PR addresses the issue.

I couldn't see any new Dockerfile in this PR just a minor change in the web's Dockerfile.
As I checked and investigated why the tmpreaper is not working with these settings, I realized that the cron service is not running on our Docker container by default and that's why I added that additional line into the `entrypoint.sh` in this [PR](https://github.com/lahmacunradio/arcsi/pull/185/files/313415bd0d46b26c508921f05aad4c212f10b7c1#diff-6f9d41d046756f0ddc2fcee0626bdb50100d12b88f293734eff742818e03efa2R2)

gammaw:
It makes all sense, silly me. Thanks for the explanation @tuz666 !

PR #176, opened 2023-06-20
Update nginx and libs

tuz666:
Updated libraries:
- babel 2.8.0 -> 2.9.1 (high severity)
- boto3 1.12.45 -> 1.21.45
- botocore 1.15.45 -> 1.24.45
- click 7.0 -> 8.1.3
- flask 1.1.1 -> 2.2.5 (high severity)
- flask-login 0.6.0 -> 0.6.2
- flask-sqlalchemy 2.4.2 -> 3.0.0
- flask-wtf 0.14.3 -> 1.0.1
- future 0.18.2 -> 0.18.3 (high severity)
- itsdangerous 1.1.0 -> 2.1.2
- jinja2 2.10.3 -> 3.1.2 (moderate severity)
- joblib 0.15.1 -> 1.2.0 (critical severity)
- mako 1.1.1 -> 1.2.2 (high severity)
- markupsafe 1.1.1 -> 2.1.1
- numpy 1.18.4 -> 1.20.0 (1 moderate and 1 high severity)
- requests 2.22.0 -> 2.31.0 (moderate severity)
- s3transfer 0.3.3 -> 0.5.2
- sqlalchemy 1.3.12 -> 1.4.18
- urllib3 1.25.8 -> 1.26.5 (1 moderate and 1 high severity)
- werkzeug 1.0.1 -> 2.2.3 (1 low and 1 high severity)

Introduced new transitive libraries:
- charset-normalizer 3.1.0
- greenlet 2.0.2
- importli-metadata 6.7.0
- zipp 3.15.0

Additional changes:
- paginate function's parameter passing changed for [flask-sqlalchemy 3.0.0 ](https://github.com/pallets-eco/flask-sqlalchemy/blob/3.0.x/CHANGES.rst#version-300)

PR #178, opened 2023-08-26
Revert nginx version update

tuz666:
In a [previous PR](https://github.com/lahmacunradio/arcsi/pull/176) I wrongly updated the version of nginx.
Although I tested the updated version locally on the prod server it had some issues and couldn't come up, I had to manually overwrite the version to the previous stable one.


gammaw (approved):
Thanks for the clean up!

PR #179, opened 2023-08-29
Prepare Tags for frontend usage

tuz666:
The goal of this PR is to extend the schemas connected to the newly introduced Tags and make them ready for the frontend usage.

Besides the schema extensions, I fixed a little glitch which caused unnecessary whitespaces, updated the Postman tests and Swagger docs.


gammaw (approved):
Did a quick visual check (no validation or deploy), looks good! Thanks for the quick action!

PR #180, opened 2023-09-04
Tag hotfix for frontend usage

tuz666:
Fixed `Tag` schemas:
- `Show` and `Item` image URLs are correct now 
- returned `Item`s are extended to contain the hosting `Show` as well

Swagger updates:
- `Tag` related API endpoint descriptions are updated
- RequestBodies are working now with multipart/form-data content
- authentication is added, protected endpoints can be used after login


gammaw (approved):
Quick review so that we can proceed. All seems good. Didn't check the testing part.

PR #181, opened 2023-09-05
Extend Tag's items and shows field with tags

tuz666:
Extend `Tag.items` and `Tag.shows` with a `tags` field


gammaw (approved):
Let's go!

Issue #182, opened 2023-10-17
Remove possible duplicate tags


tuz666:
In case of adding / editing a show or an episode, there is a chance to add duplicated tags in the form, which is not handled correctly and causes an error during the SQL insert.

PR #184, opened 2023-10-17
Remove duplicated tags

tuz666:
Fix for https://github.com/lahmacunradio/arcsi/issues/182

- New postman TCs to cover duplicated tags scenario

### DIFF
--- a/test/postman/functional_test.postman_collection.json
+++ b/test/postman/functional_test.postman_collection.json
@@ -587,7 +587,7 @@
 						},
 						{
 							"key": "taglist",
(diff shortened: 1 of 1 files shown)


gammaw (approved):
@tuz666 thanks, looks good!

A general comment for the future: I think either in the issue and/or in the linked PR we should mention what action we take to fix the bug. Right now it's only in the name of the PR and you don't mention how you technically solve it (using dictionaries via `dict` – because they by design rule out duplicates, which is by the way a very elegant fix).

PR #185, opened 2023-11-08
Archon kick off

tuz666:
Introducing a new Arcsi UI
- new Archon API endpoints, which are faster and have smaller loads
- Shows and Items Archive tabs no longer contain unnecessary info, details could be checked on the given Show's or Item's subpage.
- added a new button which links to the Swagger UI
- removed unused, legacy `/item/all` and `show/<string:show_slug>/archive` APIs

New prefixes in the API method names to make the code functionally more readable.

Hotfix for [previously wrongly merged](https://github.com/lahmacunradio/arcsi/pull/186/files#diff-6f9d41d046756f0ddc2fcee0626bdb50100d12b88f293734eff742818e03efa2) `entrypoint.sh`

Added cron job start command to `entrypoint.sh` in order to trigger tmpreaper and a finetuning in tmpreaper's configuration.

Move Swagger related files from `/arcsi/static` folder to `/arcsi/static/docs` (I tried to move it under `/docs` or `/arcsi/docs` but for me it worked just in the `/arcsi/static` folder)

Get rid of "confusing", legacy Ananasz mentions #187


arcsi/api/item.py
```diff
@@ -107,7 +115,7 @@ def list_items_latest():
 
 @arcsi.route("/item/<int:id>", methods=["GET"])
 @auth_token_required
-def view_item(id):
+def archon_view_item(id):
```
gammaw, line 118: Yeah, looks to be a good idea to distinguish the archon vs frontend usage like this.

arcsi/api/item.py
```diff
@@ -341,19 +349,19 @@ def download_play_file(id):
     return redirect(presigned, code=302)
 
 
-@arcsi.route("/item/<int:id>", methods=["DELETE"])
+@arcsi.route("/archon/item/<int:id>", methods=["DELETE"])
```
gammaw, line 352: Why not add `archon` route prefix for above archon APIs too?

arcsi/api/item.py
```diff
@@ -341,19 +349,19 @@ def download_play_file(id):
...
 @roles_required("admin")
-def delete_item(id):
+def archon_delete_item(id):
```
gammaw, line 354: Is it working now?

arcsi/api/show.py
```diff
@@ -336,9 +344,9 @@ def edit_show(id):
         )
 
 
-@arcsi.route("show/<int:id>", methods=["GET"])
+@arcsi.route("/show/<int:id>", methods=["GET"])
```
gammaw, line 347: Again, why is `/archon` prefix missing here?

entrypoint.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 python3 -m flask db upgrade
-gunicorn --worker-tmp-dir -t 30 /dev/shm -w 2 --worker-class=gthread --threads 2 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level info "arcsi:create_app('../config.py')"
+gunicorn --worker-tmp-dir /dev/shm -t 30 -w 2 --worker-class=gthread --threads 2 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level info "arcsi:create_app('../config.py')"
```
gammaw, line 3: What's this change?

gammaw (commented):
@tuz666 that's a great & badly needed change! It's gonna be such a relief for the archon users!

I added a couple of questions before I can approve, can you please check?

entrypoint.sh
tuz666: Hotfix for [previously wrongly merged](https://github.com/lahmacunradio/arcsi/pull/186/files#diff-6f9d41d046756f0ddc2fcee0626bdb50100d12b88f293734eff742818e03efa2) entrypoint.sh

arcsi/api/show.py
tuz666: As I remember this endpoint is still used by the frontend or the mobile app.
I changed just those APIs which are only used by the Arcsi itself.

arcsi/api/item.py
tuz666: Nope, I didn't add this yet, maybe in the next PR.

arcsi/api/item.py
tuz666: Same reason as below

arcsi/api/show.py
gammaw: I see. It could be mentioned in a comment on those occasions.

arcsi/api/item.py
gammaw: Also, it could be mentioned in an in-code comment that it's not working yet, just a placeholder.

gammaw (dismissed):
Thanks for the comment replies! Seems all good. Please check my replies, in case you want to add in-code comments before merging.

arcsi/api/show.py
tuz666: Sure, meanwhile I searched after the usage of this exact API in the app and frontend repo and it is safe to add the `/archon` prefix here, I added the explanation comments for the other cases.

arcsi/api/item.py
tuz666: True, I added a TODO for the DELETE APIs.

arcsi/api/item.py
tuz666: Thanks for the comments!
I double checked the API usages in the app and frontend repo, added the `/archon` prefixes into the routes where it is possible (only the Arcsi uses them itself) and marked where the legacy/old APIs are still used by the app or frontend

entrypoint.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
+service start cron
```
gammaw, line 2: @tuz666 did you read this https://github.com/lahmacunradio/arcsi/pull/177#issuecomment-1722394193? 

I still don't understand why tmpreaper should work without changing the docker compose file. I would suggest that we don't risk this (archon frontend) PR in any way (changing the entrypoint in an uncertain way is a risk) and that we experiment elsewhere. So I propose to remove this line before merging.

gammaw (changes requested):

entrypoint.sh
tuz666: @gammaw
I answered your questions under the previous PR. I wouldn't say that this change is a risk because I tested the `entrypoint.sh` change on the dev environment to make sure if it fix the tmpreaper issue or not, and as I checked this morning it worked as expected.

gammaw: Ok, thanks!

gammaw (approved):

PR #186, opened 2023-11-14
Adjust Nginx and Gunicorn config

tuz666:
**Background:**

Previously we faced a problem with our `docker-compose` setup. It was really a surprise since we didn't really do any relevant changes in the past few months.

After some investigation it turned out that the [automatic updates for Ubuntu 18.04](https://stackoverflow.com/questions/56791702/how-to-stop-disable-hold-automatic-updates-for-packages-in-linux) was enabled on our production server. The version of `go` was updated from `1.20.4` to `1.20.7` which - because of some compatibility issues ([more](https://github.com/moby/moby/issues/45935) [info](https://github.com/docker/compose/issues/11154)) - caused an `http: invalid Host header response` for every `docker-compose` commands.

Our solution was to update Ubuntu to 20.0.4 ([18.0.4 no longer supported since May 31](https://ubuntu.com/blog/18-04-end-of-standard-support)) and we started to use the latest Docker version (`24.0.7` built with go `1.20.10`) from the `apt` repository instead of the previously used Docker (`20.10.24` built with go version `1.20.7` after the automatic update) from `snap` package manager. This means that now `docker compose` is the proper command instead of the deprecated `docker-compose`.

**The problems for which this PR was made:**

With the updated environment we saw the following new Nginx and Gunicorn related failures coming up:

- `[DEBUG] Ignoring connection epipe`
- `[CRITICAL] WORKER TIMEOUT`

The purpose of this PR is to address these errors with a better configuration.
Interesting threads in this topic:

- [Frequent worker timeout](https://stackoverflow.com/questions/10855197/frequent-worker-timeout)
- [110: Connection timed out (Nginx/Gunicorn)](https://stackoverflow.com/questions/6816215/110-connection-timed-out-nginx-gunicorn)
- [Gunicorn + nginx - Ignoring EPIPE](https://stackoverflow.com/questions/55768090/gunicorn-nginx-ignoring-epipe)
- [gunicorn socket Ignoring EPIPE](https://github.com/benoitc/gunicorn/issues/1695)
- [Gunicorn Workers and Threads](https://stackoverflow.com/questions/38425620/gunicorn-workers-and-threads)

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 python3 -m flask db upgrade
-gunicorn --worker-tmp-dir /dev/shm -w 1 --worker-class=gthread --threads 4 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level info "arcsi:create_app('../config.py')"
+gunicorn --worker-tmp-dir -t 30 /dev/shm -w 2 --worker-class=gthread --threads 2 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level info "arcsi:create_app('../config.py')"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,7 +64,7 @@ jobs:
 
     - name: Run the application and Postman's tests
       run: |
-        gunicorn -b localhost:5000 -w 4 "arcsi:create_app('../config_ci.py')" --daemon --log-file aguni.log --log-level debug
+        gunicorn -b localhost:5000 -t 30 -w 2 --threads 2 "arcsi:create_app('../config_ci.py')" --daemon --log-file aguni.log --log-level debug
       env:
         FLASK_ENV: development
         UPLOAD_FOLDER: "uploads"
--- a/nginx/http_arcsi.conf.template
+++ b/nginx/http_arcsi.conf.template
@@ -12,7 +12,8 @@ server {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header Host $host;
         proxy_redirect off;
-        proxy_read_timeout 12000s;
-        proxy_connect_timeout 12000s;
+        proxy_connect_timeout 75s;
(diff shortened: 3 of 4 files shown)


gammaw (approved):
Thanks @tuz666 for the detailed description!

The code changes look all good!

A couple of important additions to your description: 

- We had two orthogonal issues: 1) our Docker version on the host machine auto-updated into a erroneous state and 2) we felt that we need to adjust our Gunicorn configuration (that determines how our Flask app is served). 
- For 1): see original comment + it turns out that `docker-compose` is deprecated and we should use `docker compose` in the future. Note that the code change in this PR doesn't mirror any of these changes (the changes have been done directly in the host machine and it affects no code in our repo). 
- For 2): the main insight comes from https://stackoverflow.com/questions/38425620/gunicorn-workers-and-threads, which is: better use 2 workers (and we also adjusted the timeouts to reasonable values).

PS:

> Docker version from the apt repository instead of the previously used Docker from snap package manager.

I think it's an issue with `docker compose` vs `docker-compose` (see above).

tuz666:
Oh sorry, I thought it will be clear that the first part about the Ubuntu and Docker changes is just a prelude or background for the later issue, which we fixed in the commit.
But anyway thanks for the comments @gammaw, I updated the description according to that.

entrypoint.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 python3 -m flask db upgrade
-gunicorn --worker-tmp-dir /dev/shm -w 1 --worker-class=gthread --threads 4 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level info "arcsi:create_app('../config.py')"
+gunicorn --worker-tmp-dir -t 30 /dev/shm -w 2 --worker-class=gthread --threads 2 -b 0.0.0.0:5666 --log-file /app/aguni.log --log-level info "arcsi:create_app('../config.py')"
```
tuz666, line 3: auch

PR #188, opened 2024-03-26
Refactor Archon (Arcsi UI) endpoints

tuz666:
Massive performance boost for Archon (Arcsi UI) pages.
- I did a refactor for the `view`/`router` endpoints:
  - Previously we called the `api`/`arcsi` endpoints inside these endpoints, which as we saw put some additional, not properly handled load on the gunicorn workers
  - I refactored these endpoints to call the exact `api`/`arcsi` functions from the code instead of calling the exposed endpoints
  - These resulted a significantly faster response for Archon show and episode subpages, and a slightly faster response for the big, 'Shows archive' and 'Items archive' pages.
  - I also had to align the flask-security related decorators, for the directly used `api`/`arcsi` endpoints/functions

I had to align the CI tests as well.

Small fix in `entrypoint.sh` for the cron job to properly start when the app container is created, this was required for tmpreaper usage.

GitHub Actions recommended to upgrade the Node version from 16 to 20.

PR was tested locally and it is also deployed currently on the dev env.
It seems like there is no frontend compatibility issues either.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
       matrix:
         postgres-version: [12]
         python-version: [3.8]
-        node-version: [16]
+        node-version: [20]
 
     services:
       postgres:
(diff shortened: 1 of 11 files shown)


gammaw (approved):
Big kudo! That was much needed and it results in a much cleaner code. Look good and logical!

PR #191, opened 2025-04-08
Update uploader handle on forms

pvj:
I didn't find any part in the views where the email value is used. 

It not quite private and useful to display uploader addresses.

This change puts user name there 

_**Note**_ When releasing this fix we should update current values for prod uploads

### DIFF
--- a/arcsi/templates/item/add.html
+++ b/arcsi/templates/item/add.html
@@ -94,7 +94,7 @@
     <label for="archive_lahmastore">Archive to Lahmastore</label>
     <input type="hidden" id="archive_lahmastore" name="archive_lahmastore" value="1"><input type="checkbox"
       onclick="this.previousElementSibling.value=this.checked?1:0" checked>
-    <input type="hidden" name="uploader" id="uploader" value="{{ current_user.email }}">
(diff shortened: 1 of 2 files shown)


pvj:
@tuz666 please review.

tuz666 (approved):
Looks good, I was just hesitating because, I didn't want to merge this before we do the DB changes for the current values.

pvj:
> Looks good, I was just hesitating because, I didn't want to merge this before we do the DB changes for the current values.

@tuz666 Valid point. As discussed this is the psql statement that you can execute previous to the merge & deploy

`UPDATE items SET uploader = u.name FROM users AS u WHERE uploader = u.email;` 

You can test it with maybe `SELECT u.name FROM users AS u, items AS e WHERE e.uploader = u.email;`

Also when running this on prod please start the session with `BEGIN;` See: https://www.postgresql.org/docs/current/tutorial-transactions.html

If the results are correct finish with command `COMMIT;` else command `ROLLBACK;`.

tuz666:
Thanks for the commit and the command!

I've just executed it and it worked as expected.

PR #195, opened 2025-04-15
Make item add and edit safer to use on client-side

pvj:
- add template, edit template updates include: better representation of episode states and action on the form
 - add, edit api: ability to data update w/o files, ability to broadcast update w/o files, extended error handling

Updated HTML templates of add and edit item. 
Refactored parts of add and edit api but it should still work with the current schema and data. This is what we have to test. 

Important changes and details include; 
 - Now you can upload au or img files by themselves
 - Now you can reset image upload or audio upload on add and edit
 - Adding new episode must include a picture too
 - If no picture added we try to fall back to the show url or the lahmacun logo
 - layout and style updates 
 - When writing files to disk: Archive to lahmastore is selectable on edit, on by default on add
 - Arcsi cleans up tmp files after successful broadcast edit

Please take a look I will be testing this a lot too.

### DIFF
--- a/arcsi/api/item.py
+++ b/arcsi/api/item.py
@@ -35,9 +35,7 @@ class ItemDetailsSchema(Schema):
     name_slug = fields.Str(dump_only=True)
     description = fields.Str()
     language = fields.Str(max=5)
-    play_date = fields.Date(
-        required=False, load_default=datetime.today() + timedelta(days=1)
-    )  # Thanks Kamil Szot!
+    play_date = fields.Date(required=True)
     image_url = fields.Str(dump_only=True)
     play_file_name = fields.Str(dump_only=True)
     live = fields.Boolean()
--- a/arcsi/templates/item/add.html
+++ b/arcsi/templates/item/add.html
@@ -67,10 +67,9 @@
     <div>
       <label for="date_info">Schedule</label>
       <p>The station schedule rotates weekly. If any broadcasting type is selected the episode airs the next time the show is broadcasted. If no episode is added during the week then the show airs the most recent episode by default.</p>
-      <p>Setting the date of airing is optional, when no date is set the default date is tomorrow - relatively, i.e. the last possible submission is the day before show airs. The schedule is reloaded each dawn.</p>
-      <p>Summary: When an episode is ready submit it within one week before scheduled broadcast time and select from the date-picker the day the show airs.</p>
-        <label for="play_date">Play date</label>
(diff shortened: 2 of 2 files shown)


pvj:
@tuz666 I fixed the code that threw test errors. Thanks for your work, it is reasssuring to have this test suite running and we can easily debug exceptions introduced to any api endpoint. 

The PR now should be ready for review. .

arcsi/api/item.py
```diff
@@ -422,17 +482,234 @@ def archon_edit_item(id):
...
+                clean_name=normalise(tag.display_name),
+            )
+            for tag in item_metadata.tags
         )
-        
+        # TODO Implement upsert https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#insert-on-conflict-upsert
         db.session.add(item)
         db.session.flush()
+        empty_files = all([not (_v) for _v in request.files.values()])
+        request_files = {k: v for k, v in request.files.items()}
```
tuz666, line 496: In my PR I totally missed this part where you separated the handling of the form depending on the existence of the attached image and audio files.
This could be a nice potential addition in the future (or maybe we can handle this under your media schema idea)

arcsi/api/item.py
```diff
@@ -422,17 +482,234 @@ def archon_edit_item(id):
...
+                broadcast_file_name=play_url_name,
+                broadcast_playlist=item.shows[0].playlist_name,
+                broadcast_show=item.shows[0].name,
+                broadcast_title=item.name,
+                image_file_name=image_url_name,
+            )
+
+            if item.airing:
+                db.session.commit()
+            # TODO cleanup tmp files after upload?
```
tuz666, line 690: Nice idea, this will help out the daily tmpreaper job, and it will also decrease the chance that the show hosts receive a storage related error.
I added this into the other PR in case of successful uploads!

arcsi/api/utils.py
```diff
@@ -122,6 +126,19 @@ def broadcast_audio(
             return True
     return False
 
+
+def delete_file(archive_base, archive_idx, archive_name):
```
tuz666, line 130: I added this function and the safe join changes into the other PR.

arcsi/api/item.py
pvj: Great!

arcsi/api/utils.py
pvj: Great!

PR #197, opened 2025-05-02
Refactoring add and edit functions

tuz666:
- Extracted the common `process_files`, `archive_files `and `broadcast_episode` methods from `add` and `edit` functions into utils.
- Introduced black formatter and formatted all `.py` files


arcsi/api/item.py
pvj, line 1: Next time you can rebase interactively -- use squash or fixup! to mold the two commits into one.

pvj (approved):
Good changes, smart choices!

PR #199, opened 2025-05-14
Improve add and edit forms

tuz666:
New add and edit templates for shows and episodes:
- Layout and style updates
- Reset button for the selected image and audio file
- Small tooltips for the Live / Archive / Broadcast

New FAQ page for show hosts which describes the basic use cases

The temp files are cleaned up in case of successful episode creation and edit

Show subpage now has a field which describes the current state of the given show's playlist (exists / empty / not-empty):
- Uploading a live episode now cleans up the playlist, it removes the previously uploaded recording to prevent starting it in the show's time slot


arcsi/api/utils.py
```diff
@@ -158,6 +158,30 @@ def broadcast_audio(
     return False
 
 
+def cleanup_show_playlist(broadcast_playlist):
+    az = AzuraArchive(
```
pvj, line 162: Did you think about putting these two to the AA object as its methods? Then you would not need to instantiate these again here.

arcsi/view/show.py
```diff
@@ -30,7 +35,8 @@ def view_show(id):
     show = archon_view_show(id)
     if hasattr(show, "status_code") and show.status_code == 404:
         return "Show not found"
-    return render_template("show/view.html", show=show)
+    empty_playlist = show_playlist_is_empty(show.playlist_name)
+    return render_template("show/view.html", show=show, empty_playlist=empty_playlist)
```
pvj, line 39: Great!!

arcsi/templates/item/edit.html
```diff
@@ -71,40 +71,51 @@ <h2>{% block title %}Edit episode item{% endblock %}</h2>
...
   <div>
-    <label for="play_file">Audio file *</label>
-    <input type="file" accept= ".mp3" name="play_file" id="play_file" value="{{ item.play_file_name }}">
+    <label for="live">Live</label>
+    <input type="hidden" name="live" id="live" 
+      {% if item.live %} value="1" {% else %} value="0" {% endif %}>
+    <input type="checkbox" onclick=
+      "this.previousElementSibling.value=this.checked?1:0; play_file.disabled=!play_file.disabled;"
+      {% if item.live %} checked {% endif %}>
+    <i class="fa fa-question-circle" aria-hidden="true"
```
pvj, line 80: Great! Let's also handle mobile somehow.

arcsi/view/show.py
```diff
@@ -35,8 +35,24 @@ def view_show(id):
...
-    return render_template("show/view.html", show=show, empty_playlist=empty_playlist)
+    az = AzuraArchive(
+        None,
+        None,
+        None,
+        None,
+        None,
+        show.get("playlist_name"),
+    )
+    existing_playlist = az.find_playlist_id()
```
pvj, line 46: This part is hard to understand in terms of what happens here. If empty_playlist is boolean then remove it and use a boolean function on the existing playlist. That's easier to change or extend later.

arcsi/api/utils.py
```diff
@@ -92,21 +99,19 @@ def find_request_params(param, default, type):
...
-):
-    if not (play_file and image_file):
+def broadcast_episode(item, image_file_name, error, error_message):
+    if not (item.play_file_name and image_file_name):
         error = True
         error_message = "ERROR: Both image and audio input are required if broadcast (Azuracast) is set"
         app.logger.debug(error_message)
     else:
         item.airing = broadcast_audio(
-            archive_base=item.shows[0].archive_lahmastore_base_url,
```
pvj, line 104: good name changes!

arcsi/view/show.py
tuz666: I moved this part into the utils layer and also renamed the functions.

arcsi/templates/item/edit.html
tuz666: I created a FAQ page which is covering the basic use cases.

arcsi/view/show.py
tuz666: Here I missed that part when the playlist is not even exists, now that's also covered :)

arcsi/templates/item/edit.html
pvj: Cool, that should be enough for starters

arcsi/view/show.py
pvj: Thank you!

pvj (dismissed):

tuz666:
Thanks for the review @pvj !
I clicked through the basic use cases on the dev environment and realized that the disabled play_file field in case of live episode upload causes an error, I removed that part and fixed the css styling (I've just copy pasted that from your PR previously...)

arcsi/api/utils.py
```diff
@@ -92,21 +99,35 @@ def find_request_params(param, default, type):
...
+    az = AzuraArchive(
+        None,
+        None,
+        None,
+        None,
+        None,
+        playlist_name,
+    )
+    playlist_exists = az.is_existing_playlist_set_playlist_id()
+    playlist_is_empty = True
```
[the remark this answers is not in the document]
pvj: (or if the function is_empty_playlist can handle situations where playlist does not exist and returns False then you can just use `return az.is_existing_play..., az.is_empty_playlist()`

PR #200, opened 2025-07-14
Update python image version from 3.8 to 3.13

tuz666:


pvj (approved):
Thanks for your work! Let us see what happens

PR #202, opened 2025-08-17
Add email configuration, enable password recovery through email

pvj:
(Note; Would bump version to 0.4.6)

What did I change?
Main changes are all done through configuration files and environment variables.  I added a link for starting the reset journey to the login_form we use. 

Why?
Users who forgot their passwords must be able to create a new password. 

How to verify?
WIth a very custom hackish setup I managed to try the whole flow on localhost and I was successfully going through the flow with arcsi UI. I will put it on staging and the following flow should work:  

(Clear session)
1. Register successfully (note your email)
2. Log out
3. Go to login form
4. Click "Forgot password" link
5. Enter registered email from step 1
6. Click the link in the email
7. Set new password

### DIFF
--- a/config.template.py
+++ b/config.template.py
@@ -19,6 +19,8 @@ SECURITY_SEND_REGISTER_EMAIL = True
 SECURITY_REGISTERABLE = True
 SECURITY_PASSWORD_SALT = "some_salt"
 SECURITY_USE_VERIFY_PASSWORD_CACHE = True
+# If mail capability is available and SMTP server is set forgot password functionality is available set this to True
+SECURITY_RECOVERABLE = False
 
 # ARCSI CONF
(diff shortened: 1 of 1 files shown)


pvj:
@tuz666 I just noticed that a couple of files need to be cleaned up, please test the version on staging first and review only afterwards. I will ping you through chat when it's deployed and configured. I should do it in 1-2 days.

pvj:
@tuz666 it is out on staging -- you can try requesting a new password with a dummy user (and real email) there.

Please check code and the service out! Thanks!

tuz666 (approved):
For some reason I couldn't see the _Forgot password_ link, except that everything seems fine, so go ahead!
We can figure this out with the next, upcoming PR.

tuz666 (approved):

PR #203, opened 2025-08-19
Small cleanups, fixes

tuz666:


arcsi/templates/show/edit.html
pvj, line 1: Oh wow! I didn't notice this either. Good catch!

arcsi/templates/show/add.html
pvj, line 1: :+1:

pvj (approved):
- Good catch on the edit fields
 - Code changes are good 
 - Nice catch with the `shpw` in components
 - Good hardening with API docs

arcsi/api/utils.py
```diff
@@ -109,10 +109,11 @@ def get_playlist_existence_and_emptiness(playlist_name):
         playlist_name,
     )
     playlist_exists = az.is_existing_playlist_set_playlist_id()
-    playlist_is_empty = True
     if playlist_exists:
```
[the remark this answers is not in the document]
pvj: @tuz666  use this

```py
return playlist_exists, az.is_empty_playlist() if playlist_exists else playlist_exists, True
```

PR #218, opened 2025-12-25
Improved item search endpoint

tuz666:
Item search endpoint now also filters episodes according to tags and show names.


baxgas (dismissed):

pvj (dismissed):
The repeated queries could be a function in api utils similar to `get_show` `get_show_with_cover` etc.

There is a question about the search params vs. the clean_name being used.  

What do you think about writing in the stagger json that it is a union (which means that this is an `OR` relation not an `AND`).

requirements.txt
```diff
@@ -1,47 +1,49 @@
-alembic==1.3.2
```
pvj: @tuz666 I thought the requirements have been upgraded already on master?

arcsi/api/utils.py
```diff
@@ -330,11 +330,36 @@ def delete_file(norms_show_name, archive_file_path):
...
+
+
+def get_show_cover_json(show_json):
+    do = DoArchive()
+    if show_json["cover_image_url"]:
+        show_json["cover_image_url"] = do.download(
+            show_json["archive_lahmastore_base_url"], show_json["cover_image_url"]
+        )
+    return show_json
+
```
pvj, line 361: I think I named the variables in a misleading way and this might have lead to this duplication. 

Are these different in any sense? Both Show objects and json (casted as dict) can handle both dot and square bracket properties. 

Maybe I am missing some context here but please take a look at this.

arcsi/api/utils.py
```diff
@@ -363,6 +388,47 @@ def get_items():
...
+    item.name_slug = normalise(item.name)
+    return item
+
+
+def get_item_fields_json(item_json, show_json):
+    do = DoArchive()
+    item_json["image_url"] = do.download(
+        show_json["archive_lahmastore_base_url"], item_json["image_url"]
+    )
+    item_json["name_slug"] = normalise(item_json["name"])
```
pvj, line 406: I am trying to review this too, something is a bit off. 

`get_item_fields_json` has no `if` predicate

There is a difference in parameters. 

It is better to keep everything as is for now but can you at least mark this with a `TODO` or `FIXME` comment? I would like to revisit this later.

arcsi/api/utils.py
```diff
@@ -363,6 +388,47 @@ def get_items():
...
+        Item.query.join(Item.shows)
+        .filter(func.lower(Show.name).contains(func.lower(param)))
+        .filter(Item.play_date < datetime.today() - timedelta(days=1))
+    )
+
+
+def search_tag_items(param):
+    return (
+        Item.query.join(Item.tags)
+        .filter(func.lower(Tag.clean_name).contains(normalise(param)))
```
pvj, line 427: In the end, is `clean_name` okay to use for search?

pvj (dismissed):
I am afraid `clean_name` might not work, this will be good to test on dev. 

Otherwise, great job!!!

arcsi/api/item.py
```diff
@@ -460,19 +461,30 @@ def frontend_search_item():
...
-            | func.lower(Item.description).contains(func.lower(param))
-        )
+    show_items = (
+        Item.query.join(Item.shows, aliased=True)
+        .filter(func.lower(Show.name).contains(func.lower(param)))
+        .filter(Item.play_date < datetime.today() - timedelta(days=1))
+    )
+    tag_items = (
+        Item.query.join(Item.tags, aliased=True)
+        .filter(func.lower(Tag.clean_name).contains(func.lower(param)))
```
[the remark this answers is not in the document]
tuz666: I changed the `func.lower(param)` to `normalize(param)`
`clean_name` is calculated as `normalise(tag.display_name)`
This way we can make sure that we won't have any problems with the special characters and diacritics.

requirements.txt
tuz666: They are in sync now.

arcsi/api/utils.py
tuz666: Both of these functions with the `_json` postfixes are used by the `/show/schedule_by` endpoint which meant to be called to query a specific day's schedule.
It filters the active shows for the given day and creates a dict/json (maybe I mixed these up) according to the `shows_scheduled_by_schema`.
Later on we overwrite the shows' `items` field with the latest, archived episode or an empty array.

arcsi/api/utils.py
tuz666: The same applies to this what I wrote above.
I double checked the frontend code and this endpoint isn't used yet.
I added a TODO comment as you suggested.

arcsi/api/utils.py
tuz666: I think it will be fine now, see the reasoning beneath your earlier comment.

arcsi/api/item.py
```diff
@@ -457,23 +445,19 @@ def archon_edit_item(id):
...
-    for item in items.items:
-        if item.image_url:
-            item.image_url = do.download(
-                item.shows[0].archive_lahmastore_base_url, item.image_url
-            )
-        item.name_slug = normalise(item.name)
-    return items_schema.dump(items.items)
+    for item in aggr_items.items:
+        get_item_fields(item)
+    return items_schema.dump(aggr_items.items)
```
pvj, line 463: Nice!

arcsi/api/show.py
```diff
@@ -105,7 +108,7 @@ def make_show(self, data, **kwargs):
 shows_schema = ShowDetailsSchema(many=True)
 shows_minimal_schema = ShowDetailsSchema(many=True, only=("id", "name"))
 shows_schedule_schema = ShowDetailsSchema(
-    many=True, exclude=("items", "contact_address")
+    many=True, exclude=("contact_address", "items", "users")
```
pvj, line 111: Good catch

arcsi/api/show.py
```diff
@@ -168,51 +171,41 @@ def archon_list_shows():
...
 def frontend_list_shows_for_schedule():
-    do = DoArchive()
     shows = Show.query.filter(Show.active == True).all()
     for show in shows:
-        if show.cover_image_url:
-            show.cover_image_url = do.download(
-                show.archive_lahmastore_base_url, show.cover_image_url
-            )
+        get_show_cover(show)
     return shows_schedule_schema.dump(shows)
```
pvj, line 179: Thank you

pvj (approved):
Looks good! The main logic is way easier to deduce thanks for putting in the effort. Did you run it on dev? Did you make any tests there?

tuz666:
Yes, I deployed both this and the frontend PR (https://github.com/lahmacunradio/frontend/pull/323) on the dev servers, and I couldn't find any problem.
Thank you very much for the review and all the suggestions!

PR #226, opened 2026-05-04
Extend filter on search endpoint

tuz666:
- Search endpoint's response should contain only archived episodes
- There is a filter on the frontend side which removes the non-archived episodes
- Restrict access to data endpoints


pvj (approved):
Yup! Good to align endpoint auth on `/data/` endpoints! 

`archived` filter is applied as well.

PR #227, opened 2026-05-17
Introduce new endpoints for radio website

tuz666:
- The radio website currently uses the /all_without_items for each show related queries
- As our plan is to extend the schedule with bi-weekly and monthly shows and our intention is to handle this feature on the backend side it is required to separate these incoming requests
- There should be at least two show related endpoints used by the website, one which returns the strictly schedule related additional fields (`week`, `day`, `start`, `end`, `frequency`, later on filtered according to the given week), and another one which returns all the shows with way less fields.


arcsi/api/item.py
```diff
@@ -461,3 +462,12 @@ def frontend_search_item():
     for item in aggr_items.items:
         get_item_fields(item)
     return items_schema.dump(aggr_items.items)
+
+
+@arcsi.route("/item/tag/<string:clean_tag>", methods=["GET"])
+@auth_token_required
```
pvj, line 468: This is because this is consumed by the frontend right?

arcsi/api/show.py
```diff
@@ -107,17 +104,30 @@ def make_show(self, data, **kwargs):
 show_partial_schema = ShowDetailsSchema(partial=True)
 shows_schema = ShowDetailsSchema(many=True)
 shows_minimal_schema = ShowDetailsSchema(many=True, only=("id", "name"))
+frontend_shows_tile_schema = ShowDetailsSchema(
```
pvj: Please rename this to something else, `frontend` means something outside of arcsi in general but could be confusing if someone thinks about the Arcsi UI views.

arcsi/api/show.py
```diff
@@ -162,12 +172,28 @@ def frontend_list_shows_without_items():
     return shows_schedule_schema.dump(get_shows_with_cover())
 
 
+@arcsi.route("/show/all_tiles", methods=["GET"])
```
pvj: Please fix this typo `all_tiles` => `all_titles`

arcsi/api/item.py
```diff
@@ -461,3 +462,12 @@ def frontend_search_item():
...
+def frontend_search_item_by_tag(clean_tag):
```
pvj, line 469: Please rename this to something more exact!

arcsi/api/utils.py
```diff
@@ -383,9 +409,9 @@ def get_managed_show(user, id):
     return managed_show
 
 
-def get_items():
-    items = Item.query.all()
-    return items
+def get_managed_shows(user):
```
pvj, line 412: 👍

arcsi/api/utils.py
```diff
@@ -435,6 +466,21 @@ def search_tag_items(param):
     )
 
 
+def search_shows_by_tag(param):
+    return Show.query.join(Show.tags).filter(
+        func.lower(Tag.clean_name).contains(func.lower(param))
```
pvj: Should we use something more strict than contains here? If we want to search _by tag_ not just _by free text_ then I think we need to check for equality. 

I'm happy to leave this as-is and postpone this change but I wanted to ask your opinion, what do you think?

pvj (changes requested):
I didn't check the tests yet -- I requested some changes because I think there is 1 naming issue. 

Great job all in all this is huge for scheduling roadmap plans!

arcsi/api/show.py
pvj: Ahhh so this is `tiles` not `titles` on purpose. I just saw the frontend pull request. 

Still please change or fix this. Think about just the url and what it does not where it is used (eg. imagine that we change tiles on frontend to grid or something like that then we either have to change Arcsi code too or leave this as-is but now the meaning is lost -- both of these make maintenance harder)

arcsi/api/utils.py
pvj: `func.lower(Tag.clean_name) == (func.lower(param))`

arcsi/api/item.py
tuz666: Yes, I created a dedicated search by tag endpoint both for episodes and shows.

arcsi/api/show.py
tuz666: I removed the `frontend` prefixes and came up with new names, for the Arcsi UI related endpoints we're using the `archon` prefixes, but that could be renamed, cleaned up later either,

arcsi/api/show.py
tuz666: I renamed this to `all_details`, I couldn't come up anything better now. But we should have a discussion about naming after we finish the schedule/frontend tasks.

arcsi/api/item.py
tuz666: If you mean the name of the method I would keep the `frontend` prefix there, that's following the current convention, even if it's not the best.
Although I changed the url to `/item/search_by_tag/<string:clean_tag>`

arcsi/api/utils.py
tuz666: I fixed this as well!

arcsi/api/item.py
pvj: Sure, this is easy to fix with a quick replace when we clean up.

arcsi/api/show.py
```diff
@@ -116,10 +116,10 @@ def make_show(self, data, **kwargs):
         "tags",
     ),
 )
-shows_schedule_schema = ShowDetailsSchema(
+shows_schedule_excluded_schema = ShowDetailsSchema(
```
pvj, line 119: I really like these names, thanks!

test/postman/functional_test.postman_collection.json
```diff
@@ -6664,7 +6664,7 @@
           }
         ],
         "url": {
-          "raw": "{{url}}/show/all_tiles",
+          "raw": "{{url}}/show/all_details",
```
pvj, line 6667: Thanks, this is great!

pvj (approved):

PR #229, opened 2026-07-05
Recaptcha for Registration

tuz666:


pvj (approved):
Thanks for the front-end work!!!

PR #232, opened 2026-07-27
New data endpoint for dynamic liquidsoap config creation

tuz666:
## Why make changes? State your reason

From programming point of view we're planning to extend our schedule for quite a long time, although the current design is not capable to return different weekly schedules.

As part of this PR my aim is to introduce `week` parameters for the `schedule` endpoints with a default value of 1, in order to remain compatible with the current frontend / mobile application implementation.

The schedule change will let us handle biweekly and monthly shows properly, and give us a bunch of time slots for possible new show ideas, but it will impact the usage of [our Liquidsoap custom configuration](https://github.com/lahmacunradio/liquidsoap/blob/master/azuracast/liquidsoap_custom.liq) as well, that should be handled in a much more flexible way.

## What has changed? Describe your work
The following endpoints were extended with an additional `week` parameter:

- `/show/all_schedule`: used by the frontend
- `/show/schedule`: used by the mobile application
- `/show/schedule_by`: not used by anyone, just aligned with the other changes

A new data endpoint (`/data/weekly_schedule`) is also created, it helps in the generation of the dynamic part (playlist [initialization](https://github.com/lahmacunradio/liquidsoap/blob/ac4f0b12c05217e7120eb658d57d80605ef21d4f/azuracast/liquidsoap_custom.liq#L87-L253) and [scheduling](https://github.com/lahmacunradio/liquidsoap/blob/ac4f0b12c05217e7120eb658d57d80605ef21d4f/azuracast/liquidsoap_custom.liq#L326-L380)) of the custom Liquidsoap configuration. It also has a `week` parameter but by default it returns the actual week's data.

Added `LIQUIDSOAP_VERSION` variable into `config.template.py`, with the help of this the possible future LiquidSoap syntax changes can be tracked.

## How to review? Add scenarios or information on the output

The PR is deployed to the dev server, the new and modified endpoints can be easily reviewed there.

## Notes. Share the github issue link, source link, or anything else

Another PR is expected to kick-off the multi week migration, there are two possible solution for this:

1. change `Show.week` Int field into a List or
2. combine the `Show.week` and `Show.frequency`


pvj (commented):
Two questions quickly that I have about the overall picture. 

  - What's the reason to have 2 separate tpl's and functions (one for init, and one for the schedule)? Is it because of schedule changes without show addition/removal are lightweight this way? 
  - Did you think about creating a Schedule class for the Liquidsoap things? Those global environment.filters are a bit weird. And also I would say, I think the file name should reflect something ore generic like `schedule.py` and then we can create the LiquidsoapSchedule class or whatever.

tuz666:
Thanks for the comments @pvj !

> * What's the reason to have 2 separate tpl's and functions (one for init, and one for the schedule)? Is it because of schedule changes without show addition/removal are lightweight this way?

I thought it would be better to handle them separately, because they have a different purpose.
Although they can be queried via the same API call, but they are separated in the response as well.

> * Did you think about creating a Schedule class for the Liquidsoap things? Those global environment.filters are a bit weird. And also I would say, I think the file name should reflect something ore generic like `schedule.py` and then we can create the LiquidsoapSchedule class or whatever.

A really good idea, generalization is always welcome!
I haven't used that much OOP in python, so let me know, if anything could be written nicer.

arcsi/api/tag.py
```diff
@@ -64,7 +62,7 @@ def make_tag(self, data, **kwargs):
 def list_tags():
     tags = Tag.query.all()
     return make_response(
-        many_tags_schema.dumps(tags),
+        tags_schema.dumps(tags),
```
pvj, line 65: Good naming changes!

arcsi/handler/schedule.py
```diff
@@ -0,0 +1,88 @@
...
+
+class LiquidsoapScheduler(Scheduler):
+
+    def __init__(self):
+        super().__init__()
+        self.config = {"liquidsoap_version": app.config["LIQUIDSOAP_VERSION"]}
+
+    def get_playlist_init_template(self):
+        playlist_init_file = open(
+            "/app/arcsi/templates/schedule/ls_playlist_init_{}.tpl".format(
```
pvj, line 46: This is just for the next time: 

if you use `with <file> as <var>:` pattern then the file doesn't have to be managed (open/close) manually. 

I will add on my media branch a file handler function for file opening, that should be the easiest way for the future

arcsi/handler/schedule.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def get_start_time(timestamp):
+        hour = Scheduler.get_hour(timestamp)
+        minute = Scheduler.get_minute(timestamp)
+        return "{}h{}m".format(hour, minute)
+
+    def get_end_time(timestamp):
+        hour = Scheduler.get_hour(timestamp)
+        minute = Scheduler.get_minute(timestamp)
+        if minute == 0:
```
pvj, line 30: Good catch!

When we touch this file next time we need to leave a comment on why the `58` and `- 2`

arcsi/api/data.py
```diff
@@ -4,11 +4,35 @@
...
+    )
+    abcd_week = week_number % 4
+    shows = (
+        Show.query.filter(Show.active == True)
+        .filter_by(week=abcd_week)
+        .order_by(Show.day, Show.start)
+        .all()
+    )
+    shows = shows_schedule_schema.dump(shows)
+    playlist_init = ls.make_playlist_init_script(shows)
```
pvj: No need to fix it now, here, we can think about this later. 

If you only need the ls.make_playlist_init and ls.make_playlist_schedule for arguments to `make_schedule_script()` then you could calculate the init and schedule results within make_schedule_script. 

F.e. this current code would become:

`ret = ls.make_schedule_script(shows)` => 

```
def make_schedule_script(show_list): 
(...)
.
.
do_stuff_here(self.make_playlist_init_script(show_list), self.make_playlist_schedule_script(show_list))
```

test/postman/robustness_test.postman_collection.json
```diff
@@ -120,107 +120,6 @@
...
-              "        expectedTimeInMilliseconds + 1,\r",
-              "        `Response came in ${pm.response.responseTime} ms`);\r",
-              "});\r",
-              "\r",
-              "const shows = JSON.parse(responseBody)\r",
-              "pm.test(\"ResponseBody fields are correct!\", function() {\r",
-              "    pm.expect(shows).to.be.an(\"array\")\r",
-              "    shows.forEach(function(show) {\r",
-              "        pm.expect(show).to.be.an(\"object\")\r",
-              "        pm.expect(show.items).to.be.an(\"array\")\r",
```
pvj, line 150: Thanks for keeping this updated!

arcsi/templates/schedule/ls_schedule_v1.tpl
```diff
@@ -0,0 +1,260 @@
...
+
+settings.server.socket.set(true)
+settings.server.socket.permissions.set(0o660)
+settings.server.socket.path.set("/var/azuracast/stations/lahmacun_radio/config/liquidsoap.sock")
+
+settings.harbor.bind_addrs.set(["0.0.0.0"])
+
+settings.tag.encodings.set(["UTF-8","ISO-8859-1"])
+settings.encoder.metadata.export.set(["artist","title","album","song"])
+
```
pvj, line 20: Just thinking out loud, maybe some of these variables could be stored in config so it is easier to change settings. 

It is a very minor case nothing serious as we don't really touch them

pvj (dismissed):
API calls look good, code is good too so I think we can move forward with this. Thanks!!

arcsi/handler/schedule.py
tuz666: nice, I won't change this one, but let's fix it as part of the media branch

arcsi/handler/schedule.py
tuz666: true that, I forgot to mention the airing of jingles, the purpose of this modification.
I've just pushed a commit with a small comment.

arcsi/api/data.py
tuz666: makes sense, I fixed this as well!

arcsi/templates/schedule/ls_schedule_v1.tpl
tuz666: I would leave this as it is now, but I agree that this should be configurable in the future.

pvj (approved):
Great job, good addition!

PR #233, opened 2026-07-27
Add stages to dockerfile

pvj:
**Why?** 
This was a huge debt, and both locally and on ci a lot of time is wasted. 

This now addresses our basic separation between infra environment and app environment setup. 

**What changed?**
Consider the previous build failing with `pip` having some dependency problems. It took more than a minute reach that stage.

```
=> [builder 2/3] RUN apt-get update     && apt-get install gcc g++     libblas-dev libffi-dev liblapack-dev libopenblas-dev libpq-dev     libjpeg-dev zlib1g-dev     musl-dev postgresql tmpreaper -y     && apt-get c  72.6s
``` 
With stages segmenting the process, any app dep changes will only take as long as the command is executed. 

```
=> CACHED [builder 2/3] RUN apt-get update     && apt-get install gcc g++     libblas-dev libffi-dev liblapack-dev libopenblas-dev libpq-dev     libjpeg-dev zlib1g-dev     musl-dev postgresql tmpreaper -y     && apt  0.0s
```

and 

```
 => [build 2/3] ADD . .                                                                                                                                                                                                   5.6s
 => [build 3/3] RUN pip3 install -r requirements.txt                                                                                                                                                                     37.2s
```

I also simplified the steps a bit around the `/app` workdir handling.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,21 @@
-FROM python:3.11.13-slim-bullseye
+FROM python:3.11.13-slim-bullseye AS builder
 
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
-
-ADD requirements.txt /app/
-WORKDIR /app
 
 RUN apt-get update \
     && apt-get install gcc g++ \
     libblas-dev libffi-dev liblapack-dev libopenblas-dev libpq-dev \
     musl-dev postgresql tmpreaper -y \
-    && apt-get clean \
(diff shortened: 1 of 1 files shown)


tuz666 (approved):
Nice, I wasn't aware of this usage of FROM, but it makes sense, thank you very much!

Btw, in the CI we have separate stages for the different tasks, and as I remember it has some caching mechanism as well, so that won't be shorten.
But it will make the prod, dev and local deployments faster for sure, thanks again!

PR #234, opened 2026-08-02
Introduce uv tool as environment and dependency management solution

pvj:
**What was the issue?**

We both wanted the requirements.txt as a stateless representation of our app's env. But also as a maintenance file to be edited and updated. 

Again a very old big huge debt. 

**What is changing?**

I added `uv` fast, simple package manager. Below find a small text, I want to add this to the maintenance docs, what do you think? Please review the text too!

To use `uv` I had to bump the base image from `bullseye` to `trixie` and python version `3.11.13`->`3.12`. Locally ran the app, no regressions.

Small changes around containers and infra. Huge change, that now `requirements.txt` is not needed anymore. I only added it here for comparison.

The current `pyproject.toml` and `uv.lock` are an exact representation for the `requirements.txt` see some background on the "migration" process below.

I hope you find it easy to use and we can adopt using it. 

--- 

# use uv package manager

## first steps
 Migrating the reqs txt to UV. I used `uv add -r requirements.txt` to get the current state of python environment under UV. Simple as that. Then some manual work was needed. I checked what the root deps are with `uv tree` and opened the `pyproject.toml` config file, and removed everything except the root packages.

### use local dev env
 0. Get `uv` python package manager
 1. Package management simply `uv add` / `uv remove`
 2. `pyproject.toml` has project deps, extras, local deps, tools
 3. Try commands first with the `--dry-run` flag!

### maintain the environment
 1. Use `uv sync`: updates the project env automatically in `pyproject.toml`.
 2. `uv audit` as sanity check.
 3. `uv lock` to produce the env's fully resolved lock-file derived from `pyproject.toml`.
 4. commit the `.python-version`, pyproject and lock files

Requirements.txt is not required anymore!

### run the project
Using UV introduces some app-related changes.
 - Container: Added to arcsi `Dockerfile` an explicit `uv lock` and the `uv`, `uvx` binaries
 - App: `entrypoint.sh` changed the system commands to `uv run...`

### more features
See `uv tool`, `uv venv`, `uv check`, `uv tree`
There is `uv version` which is not quite like the tags but could be investigated if worth to bump that as well.

---


pvj:
@tuz666 Did you try it out, what do you think?

tuz666 (approved):
Thank you very much for this change!
It really makes dependency handling easier, faster and more straightforward.

